PR #1, opened 2016-03-30
Fixed a bug where field names weren't being applied to arguments

alexkrebiehl:
Arguments for GraphQL fields were being parsed as arg0, arg1, etc.  This change will override the argument/field name if "@GraphQLName" is provided in the method parameters.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,7 @@ apply plugin: 'com.jfrog.bintray'
 apply plugin: 'maven'
 apply plugin: 'maven-publish'
 
-version = "0.5.1"
+version = "0.5.0"
 
 idea {
     project {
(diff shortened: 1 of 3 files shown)


yrashk:
Thanks for the submission! It was overlooked because I use my javac with `-parameters` option which allows me to use real parameter names (https://github.com/yrashk/graphql-java-annotations/blob/master/build.gradle#L88)

Your PR looks good, can you add a test for this? Also, can you please exclude the version change, I'll do it myself?

alexkrebiehl:
Awesome.  I added a test and reverted the version.  Thanks!

yrashk:
Thanks a lot! I will release a new version shortly (working on #2, if it will take too long, I will release before it is done)

yrashk:
Released 0.6.0 with your PR included.

PR #8, opened 2016-04-26
Fixup processing of @GraphQLConnection to not inline fields

doobry:
The definition for edge types in the relay spec is they must have fields named node and cursor, but may have additional fields related to the edge. The current call to relay.edgeType adds in all the fields of the node type, which seems incorrect, as they are present under the node field and are not loaded from it by the underlying graphql-java library.

https://facebook.github.io/relay/graphql/connections.htm

This pull uses the same method as the todomvc example from the graphql-java readme to pass in an empty list of additional fields, to remove the duplication of the fields from the node.

https://github.com/andimarek/todomvc-relay-java/blob/5c37d571a77bf88e02598f916515374fdceb5b45/src/main/java/todomvc/TodoSchema.java#L153


yrashk:
Thank you!

yrashk:
Released 0.8.1

PR #9, opened 2016-05-04
Fixed variable references not being resolved 

alexkrebiehl:
Fixed variable references not being resolved when parsing `clientMutationId` for Relay Mutations.  If the argument contains a `VariableReference` type, we find the real value from `executionContext.getVariables()`

While this works, I'm not sure if this is the best fix.  I am wondering if variable references should be resolved before reaching this stage, or if this is expected behavior...


yrashk:
Thanks a lot! Even if it's not the best fix (I am not yet sure, have to investigate the problem), as long as it works, it's better to have it than not, therefore I am merging this in.

PR #12, opened 2016-06-01
Add @GraphQLInvokeDetached for top-level methods

nmn:
When methods are defined in a class, they are always invoked on `environment.getSource()`.
Further, the `MethodDataFetcher` starts with `if (environment.getSource() == null) return null;` which means that any top level Object Type, namely `Query` and `Mutation` cannot be declared as classes with inline methods.

This pull request adds a new annotation for methods, `GraphQLInvokeDetached` that will invoke the method even if `environment.getSource` is null. This cleans up the code and saves you from writing separate DataFetcher implementations for all top level fields.

There is a small test file as well.

PS. fixed a bug where the argument names would sometimes not be picked up automatically, and instead fall back to `arg0` etc.

### DIFF
--- a/src/main/java/graphql/annotations/GraphQLAnnotations.java
+++ b/src/main/java/graphql/annotations/GraphQLAnnotations.java
@@ -389,7 +389,6 @@ public class GraphQLAnnotations {
 
     protected static GraphQLArgument argument(Parameter parameter, graphql.schema.GraphQLType t) throws IllegalAccessException, InstantiationException {
         GraphQLArgument.Builder builder = newArgument();
-        builder.name(parameter.getName());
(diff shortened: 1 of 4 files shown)


nmn:
Another possibility is to make static methods behave like this instead of the annotation. Or both?
What do you think?

yrashk:
Thanks! ❤️ I think that both is a good idea, unless it's difficult to implement.

nmn:
I'll give another PR.

BTW, how do you handle complicated Array Input types? I've got some code to fix that, but maybe I'm missing something.

yrashk:
What is the scenario/problem you need to fix, can you give me an example? (Preferably, in a separate issue)

PR #13, opened 2016-06-01
support static methods

nmn:
In continuation to the previous PR. Now you can use static methods on classes as dataFetchers that get invoked on null.


nmn:
Sorry, need to add another test.

yrashk:
Thanks!  💙

nmn:
that was fast!

yrashk:
I follow http://rfc.unprotocols.org/1/C4 — the goal is to give everybody a right to contribute and merge fast. The project hasn't adopted it formally yet, like other projects of mine [most notably https://github.com/eventsourcing/es4j] — mostly because I am lazy. Either way, my goal is to make contributors' experience awesome.

PR #20, opened 2016-09-02
change target of GraphQLType to FIELD and METHOD

luiz290788:
I had to change the target of the annotation to make it recognizable for a method or field so I could assign a specific TypeFunction for the field or method.

### DIFF
--- a/src/main/java/graphql/annotations/GraphQLType.java
+++ b/src/main/java/graphql/annotations/GraphQLType.java
@@ -19,7 +19,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
-@Target(ElementType.TYPE_USE)
+@Target({ElementType.METHOD, ElementType.FIELD})
 @Retention(RetentionPolicy.RUNTIME)
 public @interface GraphQLType {
     Class<? extends TypeFunction> value() default DefaultTypeFunction.class;


yrashk:
Hi Luiz,

I am not sure I fully understand your patch yet — can you show an example? Also, it looks like its not passing tests — does it pass them for you locally?

Yurii.

luiz290788:
I'll check the tests locally and make sure they are passing.

What I'm trying to achieve with this change is this:

``` java
@GraphQLField
@GraphQLType(MyTypeFunction.class)
public String getSomeField() {
  return someField;
}
```

Without the change I made, the GraphQLType wasn't being recognized by GraphQLAnnotations while it tried to do this:

``` java
GraphQLType annotation = method.getAnnotation(GraphQLType.class);
```

yrashk:
Gotcha. Is this https://github.com/graphql-java/graphql-java-annotations/blob/master/src/test/java/graphql/annotations/GraphQLObjectTest.java#L374-L385 what you want to achieve? Or am I missing something?

luiz290788:
That's exactly that. But the annotation wasn't being recognized.

luiz290788:
Btw, this test you just pointed is not passing. But with my change, it starts passing.

I'll make sure that all the tests are passing and will commit again in this pull request.

yrashk:
It's an interest case. I just made a similar patch:

``` diff
commit 77c140050b30ea41fa28bde66d2c429b15fad623
Author: Yurii Rashkovskii <yrashk@gmail.com>
Date:   Mon Sep 5 10:13:54 2016 -0700

    Expand @GraphQLType targets (ht @luiz290788)

diff --git a/src/main/java/graphql/annotations/GraphQLType.java b/src/main/java/graphql/annotations/GraphQLType.java
index 6342f90..15ce51e 100644
--- a/src/main/java/graphql/annotations/GraphQLType.java
+++ b/src/main/java/graphql/annotations/GraphQLType.java
@@ -19,7 +19,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;

-@Target(ElementType.TYPE_USE)
+@Target({ElementType.TYPE_USE, ElementType.METHOD, ElementType.FIELD})
 @Retention(RetentionPolicy.RUNTIME)
 public @interface GraphQLType {
     Class<? extends TypeFunction> value() default DefaultTypeFunction.class;
```

And the original test stopped passing when I run it via `gradle check`. It does pass in my IntelliJ IDEA. Before the patch, `gradle check` succeeded.

Something to figure out.

yrashk:
I think I got it, can you try using 7d083343a95734371607a853f4ffc2c55754eea9?

luiz290788:
After my patch I had to change the visibility of the class `UUIDTypeFunction` to public.

yrashk:
Let me know if the updated master is good and I will make a bugfix release later today.

luiz290788:
It's working! Thanks a lot!

Are you planning on releasing this today?

yrashk:
Yes, in a few hours

yrashk:
Released 0.11.0 on Maven Central, might take up to a few hours to become available.

Thanks for your report!

luiz290788:
Thank you for the fix!

PR #21, opened 2016-09-03
Register double as a default type for Scalars.GraphQLFloat.

ceronman:
I've noticed that `graphql-java-annotations` is not able to infer GraphQL types for `Double` and `double` fields. The GraphQL standard doesn't define a `Double` value, it only defines a `Float`. However, `graphql-java` uses `Double` as base type for `Scalars.GraphQLFloat`.

Given that `double` is a primitive type of Java, I think it makes sense to register it in `graphql-java-annotations` by default.


yrashk:
Thanks for the PR!

Issue #22, opened 2016-09-18
Using an object as argument causes IllegalArgumentException


ceronman:
When using an object as field argument, an `IllegalArgumentException` will be thrown at execution.

Example:

``` java
public class GraphQLTest {

  @GraphQLName("TestObjectInputArgument")
  public static class TestObjectInputArgument {
    @GraphQLField public String a;
    @GraphQLField public int b;
  }

  public static class TestObjectInput {
    @GraphQLField
    public TestObjectInputArgument test(@GraphQLName("args") TestObjectInputArgument args) {
      return args;
    }
  }

  public static void main(String[] args) throws Exception {
    GraphQLSchema schema =
        GraphQLSchema.newSchema().query(GraphQLAnnotations.object(TestObjectInput.class)).build();
    GraphQL graphql2 = new GraphQL(schema);
    ExecutionResult result2 =
        graphql2.execute("{ test(args:{ a:\"x\", b:1 }) { a } }", new TestObjectInput());
  }
}
```

The reason is that `MethodDataFetcher` will send a `HashMap` instead of a `TestObjectInputArgument` as the first argument of `test`.

One workaround for this to use `@GraphQLDataFetcher` and provide a custom data fetcher. However, I think that to make the API more terse, `MethodDataFetcher` should be able to handle this. One way to do it is to have a convention of making input object provide a constructor that could take a `HashMap` as argument and it's able to create the proper argument for the method.

If you think this is a good idea I could provide a PR for this.

yrashk:
Yes, let's do this. A PR will be highly appreciated!

Thank you!

Raghav2211:
@ceronman I've used your example to pass values in a method which is annotate as a @GraphQLField but it doesn't return me a result(Map) Exception is : ExecutionStrategy.resolveField(): Exception while fetching data: graphql.annotations.GraphQLAnnotationsException: Unable to find constructor


I've created a parameter constructor TestObjectInputArgument.class but it doesn't work

deventh:
@Raghav2211 yes, true, "Unable to find constructor" ... with HashMap as parameter

guy120494:
@deventh, did you use the latest version?

PR #23, opened 2016-09-18
Added support for handling input objects in MethodDataFetcher

ceronman:
This fixes issue #22 by making `MethodDataFetcher` calling a special constructor of the input object type which takes a HashMap to build the appropriate object to be passed back to the method.

This could be slightly optimized by caching some of the reflection logic in the constructor, relieving the `get` method. However, I decided to push a simpler version first.


yrashk:
Thank you!

PR #25, opened 2016-09-21
Added support for method Batched Data Fetcher

### LINKED ISSUE #24: Add support for batched data fetchers
When using a `BatchedExecutionStrategy`, `graphql-java` requires the `get` method of the data fetcher to be decorated with the `@Batched` annotation. 

Currently the only way of using batched fetchers with `graphql-java-annotations` is to use a custom `@GraphQLDataFetcher`. I think this can be solved by adding a new annotation called `GraphQLBatched` that can be added to methods and then a special subclass of `MethodDataFetcher` should internally use the `@Batched` annotation.

If you agree with this approach I can provide a PR.

ceronman:
This fixes #24.

This change allows methods to be used as Batched data fetchers. Now it's possible to write batched queries like this:

``` java
    public class SimpleBatchedField {
        @GraphQLField
        @GraphQLBatched
        public static List<String> a() {
            return Arrays.asList("one", "two");
        }
    }
```

A method with the `@GraphQLBatched` annotation should always return a list. It also has to be static, because it won't operate on a single object but a list of objects, which should be obtained from the DataFetchingEnvironment.

### DIFF
--- a/src/main/java/graphql/annotations/BatchedMethodDataFetcher.java
+++ b/src/main/java/graphql/annotations/BatchedMethodDataFetcher.java
@@ -1,0 +1,24 @@
+package graphql.annotations;
+
+import graphql.execution.batched.Batched;
+import graphql.schema.DataFetchingEnvironment;
+import lombok.SneakyThrows;
+
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+
+public class BatchedMethodDataFetcher extends MethodDataFetcher {
+    public BatchedMethodDataFetcher(Method method) {
+        super(method);
+        if (!Modifier.isStatic(method.getModifiers())) {
+            throw new IllegalArgumentException("Batched method should be static");
+        }
+    }
+
+    @SneakyThrows
+    @Batched
+    @Override
+    public Object get(DataFetchingEnvironment environment) {
+        return super.get(environment);
+    }
+}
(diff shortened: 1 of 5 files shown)


yrashk:
Excellent! Thank you very much. Can you run `gradle licenseFormat` to fix the failure?

ceronman:
Sorry, I just ran `gradle test` and didn't bother to run `gradle check`. I just added the license headers to the files!

yrashk:
No worries, thank you very much for the contribution!

lperdigaoTel:
How can I do this with a type/field class in the List instead of String?
Something like this:
class Meter{
        @GraphQLField
	@GraphQLBatched
	@GraphQLName("readings") 
	public static List<Reading> getReadings() {
		return new ArrayList<Reading>();
	}
...
}


My Reading class is something like this:
public class Reading {
        private int value;

	@GraphQLField
        @GraphQLName("value")
        @GraphQLDescription("Value of the reading")
	public int getValue() {
		return value;
	}
}


When I try to query it i get :
[ExceptionWhileDataFetching{exception=java.lang.IllegalArgumentException: object is not an instance of declaring class}]

org.eclipse.debug.core.DebugException: com.sun.jdi.ClassNotLoadedException: Type has not been loaded occurred while retrieving component type of array.

My query is something like "{meter{... readings{value}}}"
My datafetcher  just returns new Meter(); 

-----
It seems to be working somehow because if I try to run the following query
"{meter{... readings}}"
it complains that 'Sub selection required for type Reading'
 
---
If I change the batch to List<String> it works ok with a query like this " {meter{... readings }}"

PR #26, opened 2016-10-10
@GraphQLField for private field also

TKffTK:
It's now possibly to add annotations on private field, so you don't have to add them to getters / setters.
More brettyer way.

example:

public class Testtable {

  @GraphQLField
  private String name;

  public String getName() {
    return name;
  }

  public void setName(String name) {
    this.name = name;
  }
}

### DIFF
--- a/src/main/java/graphql/annotations/GraphQLAnnotations.java
+++ b/src/main/java/graphql/annotations/GraphQLAnnotations.java
@@ -200,7 +200,7 @@ public class GraphQLAnnotations {
                 builder.field(field(method));
             }
         }
-        for (Field field : object.getFields()) {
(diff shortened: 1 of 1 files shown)


yrashk:
Great, thank you! I am going to merge this, but would you mind adding a few more things in a follow up PR?
1. A test that covers this scenario
2. An improvement to this logic — getDeclaredFields() only retrieves fields of the target class but not the entire hierarchy (as was in the original behaviour), see more at http://stackoverflow.com/questions/16966629/what-is-the-difference-between-getfields-and-getdeclaredfields-in-java-reflectio

TKffTK:
Created new mergerequest about the tests and that inherid -case: #27

PR #27, opened 2016-10-11
Tests for private field and Fix for inheritation bug caused by #26

TKffTK:
Tests for private field and field inheritation. Fixes inheritation bug caused by previous merge request.


yrashk:
Thank you!

PR #28, opened 2016-10-14
private field was still broken, fixed it and added data fetch tests.

TKffTK:
I thought that it was using propertyDataGetter for fields (what tryis getter and then fallbacks to field), but it was using fieldDataGetter.
Added some logic which is trying to decide which one to use.

Basic idea on this change is that, if there is getter, use PropertyDataGetter. Otherwise use filedDataGetter directly.


yrashk:
Thank you!

Issue #39, opened 2016-12-11
Should allow for specialized constructor @GraphQLDataFetcher


crazyfrozenpenguin:
I've noticed that GraphQL will, for fields starting with "is" (e.g. isAvailable), set the DataFetcher to FieldDataFetcher as opposed to PropertyDataFetcher.

Native boolean fields starting by "isXxxx" are allowed to have a getter method of the same name as defined in the [Java Beans spec](http://download.oracle.com/otn-pub/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/beans.101.pdf?AuthParam=1481464050_3f25905d9013aa13d325e7f69c044b62) section 8.3.2:

```java
   public boolean is<PropertyName>();
```

The same is not necessarily true for Boolean object types.

Although it is discussible if such fields should be using FieldDataFetcher by default, the graphql-java-annotations should allow graphql fields to be easily overwritten by another DataFetcher basic type.

The current case generates a FieldDataFetcher by default:

```kotlin
  @ApiModelProperty(value = "")
  @GraphQLField
  var isAvailable: Boolean? = null
```

But instead I would like it to use the PropertyDataFetcher:

```kotlin
  @ApiModelProperty(value = "")
  @GraphQLField
  @GraphQLDataFetcher(PropertyDataFetcher::class)
  var isAvailable: Boolean? = null
```

And this code fails to compile because PropertyDataFetcher annotation does not have a public default constructor. To get around it I had to specialize the PropertyDataFetcher for that specific field and it can become quite repetitive having to do the same for other similar fields.

Instead, it would be nice to allow to pass argument for a given specialized constructor as in:

```java
  @GraphQLDataFetcher(value=PropertyDataFetcher::class, args="isAvailable")
  public Boolean isAvailable = null
```

yrashk:
It's an interesting idea. Mind sending a PR?

bmsantos:
GraphQL is Object Type Agnostic. Because of it there's no guarantee that the data object composition used to build the graph will be of the exact same type as the one returned from the DataFetcher.

PR #44 introduces a test that verifies such case. The TestGraphQLQuery uses the SampleDataFetcher to provide a TestSample instance. Because GraphQL does not care about class types (it cares only that the passed instance has the expected fields), the SampleDataFetcher ends-up returning a Sample instance instead of the TestSample used to create the graph. 

[...]

PR #44, opened 2016-12-13
Issue #39: Should use DataFetcher specialized constructor

bmsantos:
DataFetchers can now be instantiated using specialized constructors that accept strings for arguments.
This facilitates DataFetcher re-usability for given fields (no need to create innumerous specific types).

Issue #39


yrashk:
Thanks! (Sorry, I oversaw this before, too much context switching :)

Issue #45, opened 2016-12-22
Schema field order is non-deterministic


nyobe:
Hi there,

First, thanks for this wonderful library!

I noticed this while generating schema dumps for Relay.  The ordering of methods and fields returned by Class.getMethods and Class.getFields is [undefined](http://stackoverflow.com/questions/1097807/java-reflection-is-the-order-of-class-fields-and-methods-standardized#1097826), which ends up causing GraphQL Annotations  to generate schemas with arbitrary field orderings.

This unfortunately ends up making schema dumps look like they were modified after each execution. I was wondering if perhaps it would be preferable to sort the methods by name / signature prior to generating graphql fields so that the ordering remains stable. Thoughts?

yrashk:
It's a great idea! Mind contributing a PR?

nyobe:
Sure thing, see #46

yrashk:
Closed via #46

PR #46, opened 2016-12-23
Problem: schema field ordering is non-deterministic (#45)

nyobe:
Java reflection returns class methods and fields in undefined order

Solution: sort methods and fields by name


yrashk:
Thanks!

Issue #51, opened 2017-01-05
Argument must be annotated with @GraphQLName, otherwise it would fail


shishuwu:
If i don't add `@GraphQLName("name") `, then method would not be recognized correctly! (version: 0.13.2)

    @GraphQLField
    @GraphQLDataFetcher(MyDataFetcher.class)
    public String findByName(@GraphQLName("name") String name){
        return null;
    }

yrashk:
Do you use `-parameters` javac option?

shishuwu:
@yrashk Really appreciate, it works. 
(BTW, maybe you should update it in the document :) )

yrashk:
I agree. Can you send a PR?

PR #52, opened 2017-01-05
Fix issue #51 @GraphQLName problem

shishuwu:
Update the readme.md document for issue #51.
Argument must be annotated with @GraphQLName, otherwise it would fail


yrashk:
Thanks!

PR #60, opened 2017-02-10
EnumFunction doesn't create a new instance for every enum with the sa…

rhejtmanek:
…me typeName.

Enum's used in multiple queries were blowing up with VariableTypeMismatch because of https://github.com/graphql-java/graphql-java/issues/85 and the fact that a new GraphQLEnum instance is created every time this method was called.


yrashk:
Thank you!

PR #61, opened 2017-02-12
Support inheritence of GraphQLField annotation

chonton-elementum:
Allow GraphQLField annotation on interfaces and classes.  All methods and fields 'inherit' GraphQLField annotation value.  Add a boolean parameter to the annotation to allow turning off at individual fields or methods or child class.


yrashk:
Thank you!

PR #93, opened 2017-06-29
Use class simple name instead of name for fixing references

tdraier:
ObjectFunction.getTypeName() generates a name using getName() instead of getSimpleName(), as everywhere else - this is ok as long as it is internal to this class, but this leads to an error when a GraphQLTypeReference is created. A reference to an invalid name is created and the schema cannot be built.

### DIFF
--- a/src/main/java/graphql/annotations/DefaultTypeFunction.java
+++ b/src/main/java/graphql/annotations/DefaultTypeFunction.java
@@ -306,7 +306,7 @@ public class DefaultTypeFunction implements TypeFunction {
         @Override
         public String getTypeName(Class<?> aClass, AnnotatedType annotatedType) {
             GraphQLName name = aClass.getAnnotation(GraphQLName.class);
-            return toGraphqlName(name == null ? aClass.getName() : name.value());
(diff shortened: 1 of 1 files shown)


tdraier:
Actually, with #94 , getTypeName() method is not used anymore and could simply be removed from the TypeFunction interface, along with the buildType(String typeName, Class<?> aClass, AnnotatedType annotatedType) method. That would make the interface more clear and easier to implement.

guy120494:
Do you have any idea why there are 2 tests that fail? when I run the tests with Intellij, everything is fine, but with gradle 2 tests fail.

guy120494:
Ok, i figured it out. The problem was that when you run all the tests with maven, the types we define in different tests are not getting deleted (because we use graphqlannotations.object which is static)
But in Intellij, the tests are run separately so the types are not saved. My solution was to use graphqlannotations.getObject instead of graphqlannotations.object in the failed tests (which is not static) and all work fine.

guy120494:
I will merge this PR (until I have the time to look at #94 )

Issue #92, opened 2017-06-29
Confusions around the getInterface() method and typeRegistry


tdraier:
The GraphQLAnnotations keeps an internal registry of created objects, called "typeRegistry". However this one in only filled when calling getInterface() - calling getObject() does not register the object. Moreover , calling getInterface() on a class (not interface) returns a GraphQLObjectType and registers it, whereas calling getObject() on the same class return a similar GraphQLObjectType without registering it. Calling multiple time getObject() returns different instances, where getInterface() always return the same instance. This is quite confusing and can lead to errors (same type created multiple times).

Also the documentation of getInterface() is not correct : 

>  if its annotated with {@link GraphQLTypeResolver} it will return a {@link GraphQLObjectType} otherwise it will return a {@link GraphQLInterfaceType.Builder}.

Actually if it's annotated with GraphQLTypeResolver , it will return a GraphQLInterfaceType , and a GraphQLObjectType otherwise (never a .Builder).

In order to have a clearer api, maybe we could have the following :
- rename getInterface() to a more generic "getType()", update documentation and make it as the main entry point instead of object() (would be actually easier to use, as it would support any type)
- getObject() registers the type in the internal registry, so that multiple calls to getObject() return the same instance

Issue #90, opened 2017-06-22
Schema build fails when declaring same inputType in 2 different queries


danielocampo2:
**Context**: it has been added in graphql-java 3.0.0 a method that checks that InputTypes have to be unique and will throw an exception if this requirement isn't met.
The exception thrown is the following:
```Caused by: graphql.AssertException: All types within a GraphQL schema must have unique names. No two provided types may have the same name. No provided type may have a name which conflicts with any built in types (including Scalar and Introspection types). You have redefined the type 'inputCode' from being a 'GraphQLInputObjectType' to a 'GraphQLInputObjectType'```

Given how graphql-java-annotations works, this will prevent developers from using the same InputType as a parameter for 2 different queries, when this usecase might be totally valid.
I believe this happens because when building the schema, annotations library iterates over the parameters defined for a GraphQLField. If it is an object type (not primitive), it will always attempt to create a new InputType, even if this already was created previously, thus throwin the aforementioned exception.

**Example**: A simple example to reproduce this is to create a new class that will be used by the queries, say Code, and then declare this as parameters for 2 different queries.
```
public class Code {
  @GraphQLField
  public String firstField;
  @GraphQLField
  public String secondField;
}

public class YourQueryClass {
  @GraphQLField
  public List<Something> something(Code code) {  }
  @GraphQLField
  public List<SomethingElse> somethingElse(Code code) { }
}
```

**Possible solution:** when building the schema in the annotations library, evaluate if a class is already defined as InputType. If it is, use that, otherwise create a new one.

**Momentary Workaround:** If  the object that you defined (Code) it is not too complex, just flatten your query parameters to use primitive types (in this case, replace Code  in your queries with two String parameters for firstField and secondField)

tdraier:
Hi,
I fixed the Issue in #94, I also added your example in a test case.

PR #94, opened 2017-06-30
Keep all output types in registry, move type resolution from TypeFunctions to GraphQLAnnotations

tdraier:
Here's some change proposition related to the issue : #92  . The changes are significant but should not be breaking, api compatibility is kept and all tests pass.

Basically the idea is : 
- Keep all generated GraphQLOutputType in the GraphQLAnnotations local registry
- Enum support is added to GraphQLAnnotations
- A single method "getOutputType()" return a GraphQLOutputType for all supported stuff : objects, interfaces, unions, enum, and registers it in the registry
- Previous methods getObject() and getInterface() delegates to getOutputType() and are marked as deprecated
- ObjectFunction completely delegates to GraphQLAnnotations to create return types or references to types (also handle enums). No duplicate registry in the DefaultTypeFunction

One change was required in the tests : cleanup the registry before executing any new test, as all object types are now kept here and we have many tests defining different ObjectType with the same name.

### DIFF
--- a/src/main/java/graphql/annotations/GraphQLAnnotations.java
+++ b/src/main/java/graphql/annotations/GraphQLAnnotations.java
@@ -379,7 +379,12 @@ public class GraphQLAnnotations implements GraphQLAnnotationsProcessor {
 
         for (Class<?> iface : object.getInterfaces()) {
             if (iface.getAnnotation(GraphQLTypeResolver.class) != null) {
-                builder.withInterface((GraphQLInterfaceType) getInterface(iface));
+                String ifaceName = getTypeName(iface);
+                if (processing.contains(ifaceName)) {
+                    builder.withInterface(new GraphQLTypeReference(ifaceName));
+                } else {
+                    builder.withInterface((GraphQLInterfaceType) getInterface(iface));
+                }
                 builder.fields(getExtensionFields(iface, fieldsDefined));
             }
         }
--- a/src/main/java/graphql/annotations/GraphQLAnnotations.java
+++ b/src/main/java/graphql/annotations/GraphQLAnnotations.java
@@ -33,6 +33,7 @@ import static graphql.annotations.ReflectionKit.constructNewInstance;
 import static graphql.annotations.ReflectionKit.newInstance;
 import static graphql.annotations.util.NamingKit.toGraphqlName;
 import static graphql.schema.GraphQLArgument.newArgument;
+import static graphql.schema.GraphQLEnumType.newEnum;
 import static graphql.schema.GraphQLFieldDefinition.newFieldDefinition;
 import static graphql.schema.GraphQLInputObjectField.newInputObjectField;
 import static graphql.schema.GraphQLInterfaceType.newInterface;
@@ -50,7 +51,7 @@ public class GraphQLAnnotations implements GraphQLAnnotationsProcessor {
 
     private static final List<Class> TYPES_FOR_CONNECTION = Arrays.asList(GraphQLObjectType.class, GraphQLInterfaceType.class, GraphQLUnionType.class, GraphQLTypeReference.class);
 
-    private Map<String, graphql.schema.GraphQLType> typeRegistry = new HashMap<>();
+    private Map<String, graphql.schema.GraphQLOutputType> typeRegistry = new HashMap<>();
     private Map<Class<?>, Set<Class<?>>> extensionsTypeRegistry = new HashMap<>();
     private final Stack<String> processing = new Stack<>();
     private Relay relay = new Relay();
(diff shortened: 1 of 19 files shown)


tdraier:
I have added a new commit for also handling input types, which will  solve issue #90  .
The associated test tests the possibility to use an input type in an interface (so, the same input type is reused twice), and to also have recursion in input types. Both work by keeping input types in the same types registry.

tdraier:
Any thoughts about this PR ? It solves a lot of different issues, it would be nice to discuss about it 
thanks

SmoooZ:
I would love this PR to be merged into the master, thank you for the work @tdraier. Any chance that happens soon ?

tdraier:
Unfortunately I cannot tell you, I don't see much activity here. Hopefully we will get some news after the holidays .. ?

guy120494:
It will take time to merge it, because it will take time for me to look make a good code review.

but in general: don't you think GraphQLAnnotations is too big? how about refactor the code to make it more SOLID and readable?

tdraier:
Completely agree, the GraphQLAnnotations class could be split in smaller parts. I did not want to rewrite everything is this pull request to try to keep it understandable - at least you can see what has been changed (mainly, keeping types in a centralised repository to avoid recreating the same type many times, which already implies a lot of refactoring). I understand the review can take time, don't hesitate to ask me if some points are unclear
Thank you !

sergehuber:
I agree that it would be nice to refactor the code but the functionality in this PR is really useful. I would recommend that we first merge this and then eventually in a second phase refactor. This will also make the merging process easier.

tdraier:
Rebased to master to solve conflicts

guy120494:
Can you please fix the failing test?

src/main/java/graphql/annotations/GraphQLAnnotations.java
```diff
@@ -638,27 +693,49 @@ protected static GraphQLFieldDefinition field(Method method) throws Instantiatio
...
+                            map(field -> {
+                                GraphQLOutputType type = field.getType();
+                                GraphQLInputType inputType = getInputObject(type, newNamePrefix);
+                                return new GraphQLInputObjectField(field.getName(), field.getDescription(), inputType, null);
+                            }).
+                            collect(Collectors.toList()));
+            typeRegistry.put(inputObjectType.getName(), inputObjectType);
+            return inputObjectType;
+        } else if (graphQLType instanceof GraphQLList) {
+            return new GraphQLList(getInputObject(((GraphQLList)graphQLType).getWrappedType(), newNamePrefix));
```
yarinvak, line 731: It looks like you dont handle recursive lists, like List<List<InputType>>

yarinvak:
I have tried, using your code, to create a Query field that gets as an argument list of input type, and it throws an exception ("Introspection must provide input type for arguments.")

src/main/java/graphql/annotations/GraphQLAnnotations.java
tdraier: the getInputObject is recursive, so List of List should be handled. I've added a unit test for that and it works. Do you have a code sample that does not work as expected ?

tdraier:
"Introspection must provide input type for arguments." is an error from graphql-java - it might be that you are using at some point the same class for input and output. Since all types are stored in the registry, it's not possible anymore to use a class for both inputs and outputs - it must be different classes. So you probably got an OutputType instead of an InputType and used it as an argument, hence the exception. Do you have a code sample to  reproduce this ?

yarinvak:
```
public class InputClass {
    @GraphQLField
    public String stringInput;

    @GraphQLField
    public int intInput;

    public InputClass(LinkedHashMap<String,Object> args) {
        this.stringInput = (String) args.get("stringInput");
        this.intInput = (int) args.get("intInput");
    }
}
```

Well I have this InputClass, I only use it as an argument to a query.

While this works:
```
    @GraphQLField
    @GraphQLDataFetcher(PersonFetcher.class)
    public Person person3(@GraphQLName("inputs") InputClass inputs){return null;}

```

The following code will produce "Introspection must provide input type for arguments.":
```
    @GraphQLField
    @GraphQLDataFetcher(PersonFetcher.class)
    public Person person3(@GraphQLName("inputs") List<InputClass> inputs){return null;}

```

tdraier:
I've locally added your fields to the TestObjectList in the unit test but unfortunately did not manage to see your error. The schema is correctly created, I got this from the SchemaPrinter :

```
type TestObjectList {
   person3(inputs : InputClass) : Person
   person4(inputs : [InputClass]) : Person
   value(input : [[[InputObject]]]) : String
}
```

guy120494:
> it might be that you are using at some point the same class for input and output. Since all types are stored in the registry, it's not possible anymore to use a class for both inputs and outputs - it must be different classes

Although it is not recommended to use the same class as input and output, It is not our responsibility to enforce that

tdraier:
Actually, according to the specs ( http://facebook.github.io/graphql/October2016/#sec-Type-System ), "All types within a GraphQL schema must have unique names.". It's absolutely not possible to have an input and an output type with the same type name. Graphql-java uses internally a unique string->type map for all types, input and outputs. 
Since type name is directly derived from the class (with or without GraphQLName annotation), it's not possible to use the same class for input and output types.

yarinvak:
but since a prefix of "input" is added to input types, isnt it ok to use the same class?

guy120494:
But that's why we have the prefix "input" in `getInputObject()` method

guy120494:
The main question is "Does the mapping between java classes and graphql types is one to one?
i.e is a java class represents a single graphql type, or maybe it can represents more types?"

I think that the mapping shouldn't be one to one, so one class can represents more than one graphql type (and then we use the "input" prefix)

tdraier:
You're right, an "input" prefix was added, but I actually removed it in the branch. I don't remember why exactly (there was probably some issues to find it back in the registry), but I can try to restore it. Having the prefix solves the issue. The mapping is maybe not one class to one type, but at most it will be one class to two types.

guy120494:
So please restore it and resolve the conflicts, and I will merge it

tdraier:
I did restore the input prefix - it actually makes the code a little cleaner, especially in DefaultTypeFunction (I had a not very nice try-catch to determine if the type was input or output). However I had to replace the typeName parameter (which was actually not used) by a boolean indicating if we want an input or an output.

Issue #33, opened 2016-11-04
Arguments for a field annotated with @GraphQLConnection


laurentpellegrino:
I have a query type with a field named _jobs_:

```
@GraphQLField
@GraphQLConnection
public List<Job> jobs(JobInput jobInput) {
    ....
}
```

This last field defines an InputType in order to pass some arguments.

```
@GraphQLType
@Getter
public class JobInput {

    @GraphQLName("id")
    private long id;

    @GraphQLName("test")
    private String name;

}
```

Unfortunately, when settings the JobInput type, any GraphQL query that is submitted leads to a _NoSuchElementException_ in _MethodDataFetcher#invocationArgs_. The problem seems due to the environment that no longer contains the JobInput argument that is defined with the field. This argument is removed from the environment when _GraphQLAnnotations.ConnectionDataFetcher#get_ is invoked. There is even an explicit comment saying _Exclude arguments_. I am not sure to understand if this removal is a bug (i.e. only arguments related to relay pagination should be removed) or a voluntary action.

Is there a standard way to support arguments with a GraphQLConnection?

yrashk:
Thanks for opening this issue! I will look into this soon and will update you.

laurentpellegrino:
@yrashk Great! Let me know if there is something I can do to help.

cnole:
Any updates? I also ran into this problem. Could you fix this please?

CassandraTis:
up! Thanks!

CassandraTis:
up!

CassandraTis:
@yrashk Are there any updates on this issue? Thank you!

guy120494:
Solved

PR #95, opened 2017-07-04
Relay connections : support recursive and interfaces connection, pass arguments to DataFetcher

tdraier:
Small fixes on connections : 
- Pass arguments to connection data fetcher ( #33 )
- Support recursive and interfaces connection ( `isConnection` allows interface, union and ref)

### DIFF
--- a/src/main/java/graphql/annotations/GraphQLAnnotations.java
+++ b/src/main/java/graphql/annotations/GraphQLAnnotations.java
@@ -82,6 +82,8 @@ public class GraphQLAnnotations implements GraphQLAnnotationsProcessor {
 
     private static final Relay RELAY_TYPES = new Relay();
 
+    private static final List<Class> TYPES_FOR_CONNECTION = Arrays.asList(GraphQLObjectType.class, GraphQLInterfaceType.class, GraphQLUnionType.class, GraphQLTypeReference.class);
+
     private Map<String, graphql.schema.GraphQLType> typeRegistry = new HashMap<>();
     private final Stack<String> processing = new Stack<>();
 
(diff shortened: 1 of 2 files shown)


src/main/java/graphql/annotations/GraphQLAnnotations.java
```diff
@@ -511,7 +511,10 @@ private GraphQLOutputType getGraphQLConnection(boolean isConnection, AccessibleO
     private boolean isConnection(AccessibleObject obj, Class<?> klass, GraphQLOutputType type) {
         return obj.isAnnotationPresent(GraphQLConnection.class) &&
                 type instanceof GraphQLList &&
-                ((GraphQLList) type).getWrappedType() instanceof GraphQLObjectType;
+                (((GraphQLList) type).getWrappedType() instanceof GraphQLObjectType ||
```
guy120494: Maybe its better to initialize a set of valid types (Set <Class<?>> validTypes) and then ask if (GraphQLList) type).getWrappedType().class is in the set. (Or other type of set, maybe strings of names?)

src/main/java/graphql/annotations/GraphQLAnnotations.java
```diff
@@ -708,7 +711,9 @@ public ConnectionDataFetcher(Class<? extends Connection> connection, DataFetcher
         @Override
         public Object get(DataFetchingEnvironment environment) {
```
guy120494, line 710: Please make the ConnectionDataFetcher generic (ConnectionDataFetcher<T> implements DataFetcher<T>)

src/main/java/graphql/annotations/GraphQLAnnotations.java
```diff
@@ -708,7 +711,9 @@ public ConnectionDataFetcher(Class<? extends Connection> connection, DataFetcher
...
             // Exclude arguments
```
guy120494: Can you please delete this comment, or make it more understandable?

src/main/java/graphql/annotations/GraphQLAnnotations.java
tdraier: It could be better in terms of performances, but as the classes are not final and new sub-classes may eventually appear in graphql-java, I think it's better to keep "instanceof" tests instead of just checking if the class itself is in a list of predefined class, that will be more hard to maintain.
But instead of simple contains, we could also have the list of types in a collection and check if the type is an instance of any of them by using a lambda, could be a little nicer.
What do you think ?

src/main/java/graphql/annotations/GraphQLAnnotations.java
tdraier: I'm not sure how to do that easily, as ConnectionDataFetcher is built from the connection class returned by the annotation, which is not generic. This would imply adding generics in annotation, DispatchingConnection, and maybe other places. In all cases changing ConnectionDataFetcher is not related to this ticket.

src/main/java/graphql/annotations/GraphQLAnnotations.java
guy120494: Sounds good!

tdraier: ok, done

guy120494:
Can you make few tests please?

tdraier:
Hi,
Is there any other points/comment on this PR ?
Thank you !

guy120494:
@yarinvak any thoughts?

yarinvak:
it looks good, thanks
can you add some tests? for cases like union & interface

PR #96, opened 2017-07-04
OSGi DS 1.2 compatibility

runoncode:
At the moment the library is relying on the OSGi annotations that were introduced in the OSGi DS _1.3_ which makes it _impossible_ to deploy inside Equinox container that is widely used by many organisations.
This PR does two things:

- Introduces DS _1.2_ compatibility
- Adds _bundle_ task and _bundle.bnd_ file to package the library as OSGi bundle using Gradle and BND tool.
- Preserve binary and functional compatibility with older versions.

### DIFF
--- a/src/main/java/graphql/annotations/DefaultTypeFunction.java
+++ b/src/main/java/graphql/annotations/DefaultTypeFunction.java
@@ -22,8 +22,8 @@ import graphql.schema.GraphQLTypeReference;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
-import org.osgi.service.component.annotations.ReferencePolicyOption;
-import org.osgi.service.component.annotations.ServiceScope;
+import org.osgi.service.component.annotations.ReferenceCardinality;
+import org.osgi.service.component.annotations.ReferencePolicy;
 
 import java.lang.reflect.AnnotatedParameterizedType;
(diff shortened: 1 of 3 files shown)


danielocampo2 (approved):

src/main/java/graphql/annotations/DefaultTypeFunction.java
```diff
@@ -41,11 +41,9 @@
 import static graphql.annotations.util.NamingKit.toGraphqlName;
 import static graphql.schema.GraphQLEnumType.newEnum;
 
-@Component(scope = ServiceScope.SINGLETON, property = "type=default")
+@Component(property = "type=default")
```
runoncode, line 44: Component will remain a singleton since `serviceFactory=false` from DS 1.2 will be used.

src/main/java/graphql/annotations/DefaultTypeFunction.java
```diff
@@ -355,6 +353,18 @@ public DefaultTypeFunction() {
         typeFunctions.add(new ObjectFunction());
     }
 
+    @Reference(cardinality = ReferenceCardinality.MULTIPLE,
```
runoncode, line 356: Note: with `cardinality = ReferenceCardinality.MULTIPLE` we can use default policy option `ReferencePolicyOption.RELUCTANT`, therefore `ReferencePolicyOption.GREEDY` as was used before is not needed.

bbakerman (approved):

PR #103, opened 2017-09-13
Add support for wrapping lists with object types as input objects

jerrolatqbuzz:
Currently using an object containing list fields with other objects causes an invalid GraphQL schema, because the list item objects are not defined as input objects. This change recursively constructs input objects for objects inside lists.


guy120494:
Can you make few tests please?

yarinvak:
GraphQLList is not an implementation/extension of GraphQLObjectType so this method will never be invoked with this type of object

PR #105, opened 2017-10-09
Update to graphql-java 4.2

tdraier:
No major changes, however we could have a v3 branch to continue support on graphql 3.0


src/test/java/graphql/annotations/GraphQLObjectTest.java
```diff
@@ -622,7 +622,7 @@ public void queryOptional() {
         GraphQLObjectType object = GraphQLAnnotations.object(OptionalTest.class);
         GraphQLSchema schema = newSchema().query(object).build();
 
-        GraphQL graphQL = GraphQL.newGraphQL(schema).queryExecutionStrategy(new EnhancedExecutionStrategy()).build();
```
yarinvak, line 660: Why did you remove the queryExecutionStrategy?

yarinvak:
Looks like there are not many changes, it looks good.
By the way, version 5.0 of GraphQL-Java was released about a week ago, we should work on that too.

Can you write a little more detailed description about the parts you changed here?

src/test/java/graphql/annotations/GraphQLObjectTest.java
Nitix: He may removed it during development but forgot to put it back with its third commit. 

Also https://github.com/graphql-java/graphql-java-annotations/pull/105/files#diff-9599d6f1ee8ac402b47f46e351eead38L644 and https://github.com/graphql-java/graphql-java-annotations/pull/105/files#diff-9599d6f1ee8ac402b47f46e351eead38L675

tdraier: Yes, I restored an EnhancedExecutionStrategy for gql4.2, but did not put it back on tests where it was not needed - it's only used for relay mutations, so kept it only in RelayTest. I could restore it in these tests also, but it won't change much.

yarinvak: If its not necessary, no need

tdraier:
Here's the list of changes :

- Changed dependency to graphql-java-4.2
- Updated EnhancedStrategyExecution to inherit from AsyncSerialExecutionStrategy, updated method signatures from ExecutionStrategy (now support Future)
- Removed TypeInfoWorkaround (method is now public)
- DataFetchingEnvironment constructor changed : added missing parameters when duplicating DataFetchingEnvironment
- Cannot pass null DataFetcher in builder anymore, use fake value for building relay mutation
- MethodDataFetcher : only arguments with values are now passed as parameters, use a map with names instead of assuming all arguments in sequence

src/main/java/graphql/annotations/EnhancedExecutionStrategy.java
```diff
@@ -0,0 +1,99 @@
...
+import graphql.schema.GraphQLFieldDefinition;
+import graphql.schema.GraphQLObjectType;
+
+import java.util.HashMap;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+public class EnhancedExecutionStrategy extends AsyncSerialExecutionStrategy {
+
+    private static final String CLIENT_MUTATION_ID = "clientMutationId";
```
DoctorVoid: why do you have this const?

tdraier: Sorry I don't understand the question ? The const was not introduced in this PR but in https://github.com/graphql-java/graphql-java-annotations/commit/f39d26f , and is still used in 3 places in the class.

guy120494:
Can you please resolve the conflicts?

tdraier:
Is it ok to merge ?

Issue #71, opened 2017-03-07
GraphQL ID support ?


joel-costigliola:
It might be a dumb question but it is not clear to me whether `GraphQLID` is supported or not.
I see it defined in graphql-java [here](https://github.com/graphql-java/graphql-java/blob/ec399372416d560ac3b122d3a60f722366c5df2b/src/main/java/graphql/Scalars.java#L315) I'm not sure how it use since I use only annotations ;-)

Does it make sense to have a `GraphQLID` annotation ? Something that I could use it like:
```java
public class LogBatch {

	private final String id;

	public LogBatch(final String id) {
		this.id = id;
	}

	@GraphQLID
	public String getId() {
		return this.id;
	}

}
```

guy120494:
Fixed

PR #110, opened 2017-10-12
GraphQLID support

yarinvak:
Adds support for GraphQLID (exists in graphql-java).
Adds an annotation @GraphQLID you can put before a method/field with String/Integer/int type.
For example:
`private @GraphQLID String id;
`
It will be represented as an ID scalar.

Fixes issue https://github.com/graphql-java/graphql-java-annotations/issues/71

### DIFF
--- a/src/main/java/graphql/annotations/DefaultTypeFunction.java
+++ b/src/main/java/graphql/annotations/DefaultTypeFunction.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2016 Yurii Rashkovskii
- * <p>
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * <p>
- * http://www.apache.org/licenses/LICENSE-2.0
- * <p>
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/src/main/java/graphql/annotations/DefaultTypeFunction.java
+++ b/src/main/java/graphql/annotations/DefaultTypeFunction.java
@@ -65,7 +65,7 @@ public class DefaultTypeFunction implements TypeFunction {
         public boolean canBuildType(Class<?> aClass, AnnotatedType annotatedType) {
             return annotatedType != null
                     && (aClass == Integer.class || aClass == int.class || aClass == String.class)
-                    && (annotatedType.getAnnotation(GraphQLID.class) != null);
+                    && annotatedType.isAnnotationPresent(GraphQLID.class);
         }
 
         @Override
--- a/src/main/java/graphql/annotations/DefaultTypeFunction.java
+++ b/src/main/java/graphql/annotations/DefaultTypeFunction.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2016 Yurii Rashkovskii
- *
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
+ * <p>
  * http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


src/test/java/graphql/annotations/DefaultTypeFunctionTest.java
```diff
@@ -39,15 +57,34 @@ public void enumeration() {
...
+                        map(GraphQLEnumValueDefinition::getName).collect(Collectors.toList()),
+                Arrays.asList("someA", "B"));
         assertEquals(values.stream().
                         map(GraphQLEnumValueDefinition::getDescription).collect(Collectors.toList()),
                 Arrays.asList("a", "B"));
 
     }
 
     @Test
+    public void id() throws NoSuchMethodException, NoSuchFieldException {
```
guy120494: Meaningfull name please. What does this test do?

src/test/java/graphql/annotations/DefaultTypeFunctionTest.java
```diff
@@ -39,15 +57,34 @@ public void enumeration() {
...
+        DefaultTypeFunction instace = testedDefaultTypeFunction();
+        Method idStringMethod = DefaultTypeFunctionTest.class.getMethod("idStringMethod");
+        Method idIntegerMethod = DefaultTypeFunctionTest.class.getMethod("idIntegerMethod");
+        Method idIntMethod = DefaultTypeFunctionTest.class.getMethod("idIntMethod");
+        Field idStringField = DefaultTypeFunctionTest.class.getField("idStringField");
+        Field idIntegerField = DefaultTypeFunctionTest.class.getField("idIntegerField");
+        Field idIntField = DefaultTypeFunctionTest.class.getField("idIntField");
+
+        assertEquals(instace.buildType(idStringMethod.getReturnType(), idStringMethod.getAnnotatedReturnType()), GraphQLID);
```
guy120494: Alot of asserts. Maybe better to split to few tests

guy120494:
What happens if I put ``@GraphQLID`` on 2 fields? (2 on one class, or one in a class and one in a class that extends (type extension) the original class)

yarinvak:
If you put ` @GraphQLID `on 2 fields, they'll both be represented as IDs. I dont think there is some restriction for it on the graphql level (correct me if im wrong)

guy120494:
From the official docs: 

> ID: The ID scalar type represents a unique identifier...

So, I think there should be only 1 ID per object

guy120494:
On second thought, the uniqueness is of the value of the id, not the id itself, so it is fine by me that an object can have 2 ID's

src/main/java/graphql/annotations/DefaultTypeFunction.java
```diff
@@ -59,6 +55,25 @@ void setAnnotationsProcessor(GraphQLAnnotationsProcessor annotationsProcessor) {
...
+    private class IDFunction implements TypeFunction {
+        @Override
+        public String getTypeName(Class<?> aClass, AnnotatedType annotatedType) {
+            return Scalars.GraphQLID.getName();
+        }
+
+        @Override
+        public boolean canBuildType(Class<?> aClass, AnnotatedType annotatedType) {
+            return annotatedType != null
+                    && (aClass == Integer.class || aClass == int.class || aClass == String.class)
```
guy120494, line 67: Why do you restrict the ID to be a String or an int? What If I want my ID to be a costum type? (UUID for example)

src/main/java/graphql/annotations/DefaultTypeFunction.java
```diff
@@ -59,6 +55,25 @@ void setAnnotationsProcessor(GraphQLAnnotationsProcessor annotationsProcessor) {
...
+                    && (annotatedType.getAnnotation(GraphQLID.class) != null);
```
Nitix: You shouldn't use `getAnnotation(GraphQLID.class)` to verify that the annotation is present. It may throw a NullPointerExeception. You should use instead `isAnnotationPresent(GraphQLID.class)`.

Nitix (changes requested):

src/main/java/graphql/annotations/DefaultTypeFunction.java
yarinvak: Because the spec defines GraphQLID to be integer/string, it can only accept those 2 types:

> When expected as an input type, any string (such as "4") or integer (such as 4) input value should be coerced to ID as appropriate for the ID formats a given GraphQL server expects. Any other input value, including float input values (such as 4.0), must raise a query error indicating an incorrect type.

src/main/java/graphql/annotations/DefaultTypeFunction.java
yarinvak: You're right, thanx :)

PR #111, opened 2017-10-12
Fixes Null / NonNull on GraphQLConnection

tdraier:
This fixes the following issues : 
- When a field/method has GraphQLNonNull and GraphQLConnection , GraphQLConnection is ignored
- When a field/method has NotNull and GraphQLConnection , NotNull is ignored
- Returning null on a field annotated with GraphQLConnection crashes (with default DispatchingConnection )

By the way, I don't clearly understand in which cases GraphQLNonNull should be used in favour of NotNull. Looks like NotNull can be used on field/method and arguments, and GraphQLNonNull can be used on field/method (actually, on return types) and classes - so field and methods can actually have NotNull and GraphQLNonNull (and having both will result in a field wrapped twice). What is the prefered way to declare a non-null field ?

### DIFF
--- a/src/main/java/graphql/annotations/GraphQLAnnotations.java
+++ b/src/main/java/graphql/annotations/GraphQLAnnotations.java
@@ -582,15 +582,13 @@ public class GraphQLAnnotations implements GraphQLAnnotationsProcessor {
             type = (GraphQLOutputType) ((GraphQLNonNull) type).getWrappedType();
         }
 
-        if (type instanceof GraphQLList) {
-            graphql.schema.GraphQLType wrappedType = ((GraphQLList) type).getWrappedType();
-            assert wrappedType instanceof GraphQLObjectType;
-            String annValue = field.getAnnotation(GraphQLConnection.class).name();
-            String connectionName = annValue.isEmpty() ? wrappedType.getName() : annValue;
-            GraphQLObjectType edgeType = relay.edgeType(connectionName, (GraphQLOutputType) wrappedType, null, Collections.<GraphQLFieldDefinition>emptyList());
-            type = relay.connectionType(connectionName, edgeType, Collections.emptyList());
-            builder.argument(relay.getConnectionFieldArguments());
-        }
+        graphql.schema.GraphQLType wrappedType = ((GraphQLList) type).getWrappedType();
+        assert wrappedType instanceof GraphQLObjectType;
+        String annValue = field.getAnnotation(GraphQLConnection.class).name();
+        String connectionName = annValue.isEmpty() ? wrappedType.getName() : annValue;
+        GraphQLObjectType edgeType = relay.edgeType(connectionName, (GraphQLOutputType) wrappedType, null, Collections.<GraphQLFieldDefinition>emptyList());
+        type = relay.connectionType(connectionName, edgeType, Collections.emptyList());
+        builder.argument(relay.getConnectionFieldArguments());
 
         if (isNonNull) {
             type = new GraphQLNonNull(type);
--- a/src/main/java/graphql/annotations/GraphQLAnnotations.java
+++ b/src/main/java/graphql/annotations/GraphQLAnnotations.java
@@ -471,7 +471,7 @@ public class GraphQLAnnotations implements GraphQLAnnotationsProcessor {
         }
 
         GraphQLOutputType outputType = (GraphQLOutputType) typeFunction.buildType(field.getType(), field.getAnnotatedType());
-        outputType = field.getAnnotation(NotNull.class) == null ? outputType : new GraphQLNonNull(outputType);
+        outputType = field.isAnnotationPresent(NotNull.class) ? new GraphQLNonNull(outputType) : outputType;
 
         boolean isConnection = isConnection(field, outputType);
         if (isConnection) {
@@ -577,22 +577,22 @@ public class GraphQLAnnotations implements GraphQLAnnotationsProcessor {
     }
 
     private GraphQLOutputType getGraphQLConnection(AccessibleObject field, GraphQLOutputType type, GraphQLFieldDefinition.Builder builder) {
-        boolean isNonNull = (type instanceof GraphQLNonNull);
-        if (isNonNull) {
+        if (type instanceof GraphQLNonNull) {
             type = (GraphQLOutputType) ((GraphQLNonNull) type).getWrappedType();
+            return new GraphQLNonNull(internalGetGraphQLConnection(field, type, builder));
+        } else {
+            return internalGetGraphQLConnection(field, type, builder);
         }
+    }
 
+    private GraphQLOutputType internalGetGraphQLConnection(AccessibleObject field, GraphQLOutputType type, GraphQLFieldDefinition.Builder builder) {
         graphql.schema.GraphQLType wrappedType = ((GraphQLList) type).getWrappedType();
         assert wrappedType instanceof GraphQLObjectType;
-        String annValue = field.getAnnotation(GraphQLConnection.class).name();
-        String connectionName = annValue.isEmpty() ? wrappedType.getName() : annValue;
+        String connectionName = field.getAnnotation(GraphQLConnection.class).name();
+        connectionName = connectionName.isEmpty() ? wrappedType.getName() : connectionName;
         GraphQLObjectType edgeType = relay.edgeType(connectionName, (GraphQLOutputType) wrappedType, null, Collections.<GraphQLFieldDefinition>emptyList());
         type = relay.connectionType(connectionName, edgeType, Collections.emptyList());
         builder.argument(relay.getConnectionFieldArguments());
-
-        if (isNonNull) {
-            type = new GraphQLNonNull(type);
-        }
         return type;
     }
 
--- a/src/main/java/graphql/annotations/GraphQLAnnotations.java
+++ b/src/main/java/graphql/annotations/GraphQLAnnotations.java
@@ -21,7 +21,6 @@ import graphql.schema.GraphQLNonNull;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
 
-import javax.validation.constraints.NotNull;
 import java.lang.reflect.*;
 import java.util.*;
 import java.util.function.Function;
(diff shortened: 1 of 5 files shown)


src/main/java/graphql/annotations/GraphQLAnnotations.java
```diff
@@ -495,24 +496,35 @@ private boolean checkIfPrefixGetterExists(Class c, String prefix, String propert
...
+        }
+        return type;
     }
 
-    private boolean isConnection(AccessibleObject obj, Class<?> klass, GraphQLOutputType type) {
+    private boolean isConnection(AccessibleObject obj, GraphQLOutputType type) {
+        if (type instanceof GraphQLNonNull) {
+            type = (GraphQLOutputType) ((GraphQLNonNull) type).getWrappedType();
+        }
+        final GraphQLOutputType wrappedType = type;
```
guy120494: Can you rename wrappedType to actualType? It is not really a wrapped type (it is probably a list)
I dont like the line ``aClass -> aClass.isInstance(((GraphQLList) wrappedType).getWrappedType())``

src/main/java/graphql/annotations/GraphQLAnnotations.java
```diff
@@ -495,24 +496,35 @@ private boolean checkIfPrefixGetterExists(Class c, String prefix, String propert
...
-                outputType = RELAY_TYPES.connectionType(connectionName, edgeType, Collections.emptyList());
-                builder.argument(RELAY_TYPES.getConnectionFieldArguments());
-            }
+    private GraphQLOutputType getGraphQLConnection(AccessibleObject field, GraphQLOutputType type, GraphQLFieldDefinition.Builder builder) {
+        boolean isNonNull = (type instanceof GraphQLNonNull);
+        if (isNonNull) {
+            type = (GraphQLOutputType) ((GraphQLNonNull) type).getWrappedType();
+        }
+
+        if (type instanceof GraphQLList) {
```
guy120494: Do we need that check? We assume that ``type`` is already a list here (We check it in ``isConnection`` function)

tdraier: No, we absolutely don't need it, it was there and did not remove it but I agree it's useless

src/main/java/graphql/annotations/GraphQLAnnotations.java
tdraier: Do you prefer : 
```
        if (obj.isAnnotationPresent(GraphQLConnection.class) && actualType instanceof GraphQLList) {
            graphql.schema.GraphQLType wrappedType = ((GraphQLList) actualType).getWrappedType();
            return TYPES_FOR_CONNECTION.stream().anyMatch(aClass -> aClass.isInstance(wrappedType));
        }
        return false;
```
It's a little more verbose , buy maybe better ?

guy120494:
Maybe we should restrict the NonNull only for arguments, and then we will have a logical seperation.
GraphQLNonNull is on output types, and NonNull is on input types.
If we could have only 1 annotations it would be great, but I dont know how easy it is to use the same annotation to do 2 different things

tdraier:
I think it will be ok to use only NotNull on arguments, so ignore it on fields/methods. If we prefer to have one single annotation it's also not an issue to use GraphQLNonNull on arguments (but hardly NotNull, which does not support enough targets). 

I actually don't like the fact that the DefaultTypeFunction wraps the type with GraphQLNonNull. Either the annotation is on the class itself - and then the getObject(class) should directly return a GraphQLNonNull, or the annotation is on the type usage (return type) - and the GraphQLNonNull is added in getField(). It should not be the role of DefaultTypeFunction to do this.

What do you think ?

src/main/java/graphql/annotations/GraphQLAnnotations.java
```diff
@@ -390,12 +390,13 @@ protected GraphQLFieldDefinition getField(Field field) throws GraphQLAnnotations
             typeFunction = newInstance(annotation.value());
         }
 
-        GraphQLOutputType type = (GraphQLOutputType) typeFunction.buildType(field.getType(), field.getAnnotatedType());
+        GraphQLOutputType outputType = (GraphQLOutputType) typeFunction.buildType(field.getType(), field.getAnnotatedType());
+        outputType = field.getAnnotation(NotNull.class) == null ? outputType : new GraphQLNonNull(outputType);
```
yarinvak: can you use isAnnotationPresent instead of getAnnotation?

src/main/java/graphql/annotations/GraphQLAnnotations.java
```diff
@@ -495,24 +496,33 @@ private boolean checkIfPrefixGetterExists(Class c, String prefix, String propert
...
-            }
+    private GraphQLOutputType getGraphQLConnection(AccessibleObject field, GraphQLOutputType type, GraphQLFieldDefinition.Builder builder) {
+        boolean isNonNull = (type instanceof GraphQLNonNull);
+        if (isNonNull) {
+            type = (GraphQLOutputType) ((GraphQLNonNull) type).getWrappedType();
+        }
+
+        graphql.schema.GraphQLType wrappedType = ((GraphQLList) type).getWrappedType();
+        assert wrappedType instanceof GraphQLObjectType;
+        String annValue = field.getAnnotation(GraphQLConnection.class).name();
```
yarinvak: I would call it connectionName, and instead of initializing a new variable next line, would use it.

src/main/java/graphql/annotations/GraphQLAnnotations.java
```diff
@@ -495,24 +496,33 @@ private boolean checkIfPrefixGetterExists(Class c, String prefix, String propert
...
-                GraphQLObjectType edgeType = RELAY_TYPES.edgeType(connectionName, (GraphQLOutputType) wrappedType, null, Collections.<GraphQLFieldDefinition>emptyList());
-                outputType = RELAY_TYPES.connectionType(connectionName, edgeType, Collections.emptyList());
-                builder.argument(RELAY_TYPES.getConnectionFieldArguments());
-            }
+    private GraphQLOutputType getGraphQLConnection(AccessibleObject field, GraphQLOutputType type, GraphQLFieldDefinition.Builder builder) {
+        boolean isNonNull = (type instanceof GraphQLNonNull);
+        if (isNonNull) {
+            type = (GraphQLOutputType) ((GraphQLNonNull) type).getWrappedType();
+        }
+
```
yarinvak: What about extracting this code to another method (this method is pretty long), and then calling the extracted method inside the if block (Two occurences of this method- one in the if block and one outside it, in case its not NonNull).
(Cause I didnt like the two occurences of the same if condition in the beggining and the end of the method)

yarinvak (changes requested):

yarinvak:
> I actually don't like the fact that the DefaultTypeFunction wraps the type with GraphQLNonNull. Either the annotation is on the class itself - and then the getObject(class) should directly return a GraphQLNonNull, or the annotation is on the type usage (return type) - and the GraphQLNonNull is added in getField(). It should not be the role of DefaultTypeFunction to do this.
> 
> What do you think ?
> 

It sounds right if there is no other case which the buildType should wrap the type with GraphQLNonNull.

guy120494:
So I think it is better to stick with one annotation, so lets go with GraphQLNonNull
By the way - what does it mean to put it on a class?

src/main/java/graphql/annotations/GraphQLAnnotations.java
tdraier: done

src/main/java/graphql/annotations/GraphQLAnnotations.java
tdraier: done

src/main/java/graphql/annotations/GraphQLAnnotations.java
tdraier: agree, it makes this cleaner

tdraier:
Having the annotation GraphQLNonNull on the class will make all fields returning this class wrapped by a GraphQLNonNull, even if there's no annotation on the field itself. However, getting the type with getObject() won't return a type wrapped with GraphQLNonNull. Actually I think it's quite confusing, and we could also get rid of that.
Should I create another PR for these changes, or do that here ?

guy120494:
You can do that here

yarinvak (approved):

guy120494:
There is still use of NotNull annotation, but we agree to remove it in favor of GraphQLNotNull

src/main/java/graphql/annotations/GraphQLAnnotations.java
```diff
@@ -618,11 +628,13 @@ protected GraphQLFieldDefinition getField(Method method) throws GraphQLAnnotatio
             outputTypeFunction = typeFunction;
         }
 
-        GraphQLOutputType type = (GraphQLOutputType) outputTypeFunction.buildType(method.getReturnType(), annotatedReturnType);
-        GraphQLOutputType outputType = method.getAnnotation(NotNull.class) == null ? type : new GraphQLNonNull(type);
+        GraphQLOutputType outputType = (GraphQLOutputType) outputTypeFunction.buildType(method.getReturnType(), annotatedReturnType);
+        outputType = method.getAnnotation(NotNull.class) == null ? outputType : new GraphQLNonNull(outputType);
```
guy120494: 1. Why NotNull and not GraphQLNotNull?
2. Please use isAnnotationsPresent

tdraier: I did not commit it yet :-)

guy120494: ohh, sorry. I thought you were because there are not conflicts now

tdraier:
Ok, I finally kept the code in DefaultTypeFunction instead of GraphQLAnnotations - the code is centralized in one single place, and recursive calls to buildType() will take the annotations into account.  
I completely removed support for NotNull - which is much more clear, as GraphQLNonNull can be used anywhere, where NotNull was only possible on field and parameters, but not on type usages ( like List<@GraphQLNonNull String> ). It's also not possible anymore to set the annotation on the class itself, since it was somewhat confusing. Now you can have something like :

`public @GraphQLNonNull String test(@GraphQLNonNull List<@GraphQLNonNull String> inputs)`

PR #113, opened 2017-10-16
Type extensions

tdraier:
- Added documentation
- Also added public methods in GraphQLAnnotationsProcessor interface


README.md
```diff
@@ -113,6 +113,77 @@ field.
...
+}
+```
+
+```
+@GraphQLTypeExtension(Human.class)
+public class HumanExtension {
+    @GraphQLField
+    public String homePlanet() { }
+}
+```
```
guy120494, line 158: Can you add here the part where you register the extension?

PR #114, opened 2017-10-19
Avoid duplicate definitions of the same types when using connections

tdraier:
The issue happens when two different type want to use the same type of connection. The connection/edge type is created multiple times, and conflicts when the schema is created. It also happen if an interface defines a connection field - each implementation will recreate an output type for the connection/edge. Simply storing the type in the typeRegistry and reusing it fixes the issue.

### DIFF
--- a/src/main/java/graphql/annotations/GraphQLAnnotations.java
+++ b/src/main/java/graphql/annotations/GraphQLAnnotations.java
@@ -473,7 +473,8 @@ public class GraphQLAnnotations implements GraphQLAnnotationsProcessor {
 
         boolean isConnection = isConnection(field, outputType);
         if (isConnection) {
-            outputType = getGraphQLConnection(field, outputType, builder);
+            outputType = getGraphQLConnection(field, outputType);
+            builder.argument(relay.getConnectionFieldArguments());
         }
 
         builder.type(outputType);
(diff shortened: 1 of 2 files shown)


src/test/java/graphql/annotations/GraphQLConnectionTest.java
```diff
@@ -270,4 +289,11 @@ public void customConnection() {
         assertEquals(edges.size(), 0);
     }
 
+    @Test
+    public void duplicateConnection() {
+        GraphQLObjectType object = GraphQLAnnotations.object(DuplicateTest.class);
+        GraphQLSchema schema = newSchema().query(object).build();
+        System.out.println("ok");
```
guy120494: Please delete this line

src/main/java/graphql/annotations/GraphQLAnnotations.java
```diff
@@ -588,12 +588,21 @@ private GraphQLOutputType internalGetGraphQLConnection(AccessibleObject field, G
         assert wrappedType instanceof GraphQLObjectType;
```
guy120494: I know it is not relevant for this PR, but why does this assert make sense? we allow other types (interface,enum,union etc.)

src/main/java/graphql/annotations/GraphQLAnnotations.java
```diff
@@ -588,12 +588,21 @@ private GraphQLOutputType internalGetGraphQLConnection(AccessibleObject field, G
...
         String connectionName = field.getAnnotation(GraphQLConnection.class).name();
         connectionName = connectionName.isEmpty() ? wrappedType.getName() : connectionName;
-        GraphQLObjectType edgeType = relay.edgeType(connectionName, (GraphQLOutputType) wrappedType, null, Collections.<GraphQLFieldDefinition>emptyList());
-        type = relay.connectionType(connectionName, edgeType, Collections.emptyList());
+        GraphQLObjectType edgeType = ensureTypeUniqueness(relay.edgeType(connectionName, (GraphQLOutputType) wrappedType, null, Collections.<GraphQLFieldDefinition>emptyList()));
```
guy120494: ensure suggests that the function is boolean. Please rename it to something better
(like getUniqueType, or getActualType)

src/main/java/graphql/annotations/GraphQLAnnotations.java
```diff
@@ -588,12 +588,21 @@ private GraphQLOutputType internalGetGraphQLConnection(AccessibleObject field, G
...
+        type = ensureTypeUniqueness(relay.connectionType(connectionName, edgeType, Collections.emptyList()));
         builder.argument(relay.getConnectionFieldArguments());
         return type;
     }
 
+    private GraphQLObjectType ensureTypeUniqueness(GraphQLObjectType type) {
```
guy120494: ensure suggests that the function is boolean. Please rename it to something better
(like getUniqueType, or getActualType)

src/main/java/graphql/annotations/GraphQLAnnotations.java
tdraier: Yes, completely agree, the assert does not make sense. The wrapped type is an OutputType ( actually, a type from TYPES_FOR_CONNECTION ). I tried to simplify the code by being more specific on the types. I also moved out the operation on the builder, as I don't think a `getGraphQLConnection()` method should mutate something, in my opinion it's better to keep it outside.

tdraier:
Still in the connections, I find it very confusing that the graphql.annotations.Connection class is actually a DataFetcher that returns a graphql.relay.Connection . What would you think of renaming it to something like `ConnectionDataFetcher`  ?

guy120494:
You are right. The connection has a bigger issue: it doenst make a real paging.
The connection pulls all the data from the DB, so it losing its point of being a paging.

Me, @yarinvak and Imry koch (which I cant tag for some reason) are working on a PR to fix that,
and in the PR we are doing some factoring to make the code more readable

tdraier:
Great news, it will be nice to have pagination here ! So I will let you do the refactoring/renaming to make the code more clear

PR #115, opened 2017-10-19
Removed wrapper to fix usage of SchemaPrinter

tdraier:
I had issues using the SchemaPrinter because of the GraphQLObjectTypeWrapper class. SchemaPrinter is very useful but only supports the standard types provided by graphql-java - having a "GraphQLObjectTypeWrapper" type breaks the tool. 
Actually I was wondering what was the usage of GraphQLObjectTypeWrapper - it seems to be used only in unit test to get the class back from the type. If the feature is really needed, I think it would be better to keep the GraphQLObjectType and add a method in GraphQLAnnotation like `getBackingClass(GraphQLType)`. 
By the way, the wrapper only exists for GraphQLObjectType, but if it's available for this type it should also be possible to get the backing class for interface, union, input, etc... having a single method looking in the types registry would be more transparent than adding wrappers around all types.


guy120494:
As for now, I don't see why we need this method, so Im merging this PR.
If at some point in the future we see we need that, we would make the method

PR #121, opened 2017-10-25
changed groupid

guy120494:
changed the groupid.
In Maven the groupid is io.github.graphql-java from now on
In Bintray the groupid is the default that bintray sets from now on

@andimarek 
@bbakerman 

Please approve this if you are OK with is (or give me a comment to fix it)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -85,7 +85,7 @@ publishing {
     publications {
         maven(MavenPublication) {
             from components.java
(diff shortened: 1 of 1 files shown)


andimarek (approved):
I am fine with that. Not sure how it will work with maven central, you will have to try.

bbakerman (approved):

PR #129, opened 2017-11-21
Hotfix - check if the GraphQL type has extension annotation

yarinvak:
I found a bug - we could not map between two entities (DB entity and GraphQL entity) because the ExtensionDataFetcherWrapper tried to cast the source to the GraphQL entity where it should not do that - so we check now if the class is an extension and only then use the ExtensionDataFetcherWrapper

### DIFF
--- a/src/test/java/graphql/annotations/GraphQLObjectTest.java
+++ b/src/test/java/graphql/annotations/GraphQLObjectTest.java
@@ -133,6 +133,43 @@ public class GraphQLObjectTest {
         }
     }
 
+    public static class TestMappedObject {
+        @GraphQLField
+        public String name;
+    }
+
+    public static class TestObjectDB{
+        public String name;
+
+        public TestObjectDB(String name) {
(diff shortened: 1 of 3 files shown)


imrikoch1212:
can you add some tests please?

src/test/java/graphql/annotations/GraphQLObjectTest.java
```diff
@@ -214,9 +214,6 @@ public void fields() {
         assertEquals(fields.get(5).getName(), "privateTest");
         assertEquals(fields.get(6).getName(), "publicTest");
 
-        assertEquals(fields.get(5).getDataFetcher().getClass(), ExtensionDataFetcherWrapper.class);
-        assertEquals(fields.get(6).getDataFetcher().getClass(), ExtensionDataFetcherWrapper.class);
```
imrikoch1212, line 218: maybe instead of delete this assert you can check which class is it

PR #133, opened 2017-11-27
Upgrade to graphql-java 6.0

jplock:


guy120494:
Graphql java 6 has a few API improvements (https://github.com/graphql-java/graphql-java/releases).
We need to if we need to adapt our code before we merge

tdraier:
+1 , required to use with graphql-java-servlet which now depends on graphql-java 6 .

sergehuber:
Would be great if this could be merged, especially since the change is simple.

PR #141, opened 2017-12-27
FEATURE: Properly support Spring Boot Devtools

skurfuerst:
(see inline comments for a full explanation)


guy120494:
Please add the license for the build to pass

PR #144, opened 2018-01-18
Add option to use Asynchronous loading of connection data.

vinayknl:
Using CompletableFuture so that Paginated Data loading can use Asynchronous loading.


yarinvak:
Is it conflicts if someone doesn't want to use it asynchronious?

vinayknl:
@yarinvak It makes sense to keep the original behavior as it is and add asyn as optional. So, added a wrapper and an option to set async in the annotation (defaults to false). Please review again.
Thanks,
Vinay

guy120494:
Please add some tests to make sure that `AsyncConnectionDataFetcher` works (you can have the same tests as the regular `ConnectionDataFetcher`, just to see that it does what it is supposed to do)

vinayknl:
I have added a test to ensure both sync and async way fetches data. I am not sure if we need a way to check if this is really async. Please check and let me know if any additional tests are desirable.

PR #151, opened 2018-02-06
Allows to defines a getInstance() method on TypeResolver/DataFetcher/…

tdraier:
…Validators/DefaultValues..

Small improvement (last one !) : when calling newInstance(), it first checks if a getInstance() returning a singleton is available and call it if possible. This is very useful for TypeResolver, which I need to instantiate on my side (to correctly fill the available types), and share the same instance with graphql-java (otherwise, I need to use ugly static maps of types). It's also available and can be useful for all types which are created using this method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,17 +34,25 @@ GraphQLObjectType object = GraphQLAnnotations.object(SomeObject.class);
 
 ## Defining Interfaces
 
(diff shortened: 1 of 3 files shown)


guy120494:
Can you add something about that in the readme?

PR #159, opened 2018-03-20
Added support for arrays

### LINKED ISSUE #158: Support arrays
Arrays are not supported. I think we should support arrays by making them GraphQLList

osher-sade:
Fix #158


src/test/java/graphql/annotations/GraphQLIterableTest.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    public static class ListFetcher implements DataFetcher<List<TestObjectDB>> {
+
+        @Override
+        public List<TestObjectDB> get(DataFetchingEnvironment environment) {
+            return Collections.singletonList(new TestObjectDB("test", "test"));
+        }
+    }
+
```
yarinvak: I would add test that checks an array in array
and array in an array objects
Just to be completely sure that everything works as expected

src/main/java/graphql/annotations/processor/typeFunctions/IterableFunction.java
```diff
@@ -18,39 +18,48 @@
...
 import graphql.schema.GraphQLType;
 
+import java.lang.reflect.AnnotatedArrayType;
 import java.lang.reflect.AnnotatedParameterizedType;
 import java.lang.reflect.AnnotatedType;
 import java.lang.reflect.ParameterizedType;
 
 /**
- * Support for the Iterable things like Lists / Sets / Collections and so on..
+ * Support for the Iterable things like Lists / Sets / Collections / Arrays and so on..
```
guy120494: Array is not iterable, so add a new line describing that array is also supported

PR #160, opened 2018-03-20
Simple paging

guy120494:
I have added a new feature for paging.
The relay paging is very nice, but sometime it is overkill.
This new feature is simple paging, meaning you don't have cursors, edges and nodes, only
the arguments (first,last,after,before) and then you directly specify the fields of your entity
The cursors are just numbers, and the fetcher can fetch according to the DB row number.
It is also convenient when the data is sorted according to the DB rows' numbers.

To use it, one needs to annotate his field with `@GraphQLSimpleConnection` and the return type must be an instance of `SimplePaginatedData`. There is a class named `SimplePaginatedDataImpl` to make life easier.

`SimplePaginatedData` has two methods:
getData  - get the data of the relevant entities
getOverAllCount - get the amount of all the entities that can be retrieved

The query for this simple connection would look like that:
```
getHeroesSimpleConnection(first:2, aftet:10){
     overAllCount
     data{
         name
         mana
         hp
      }
}
```


src/main/java/graphql/annotations/connection/AbstractPaginatedData.java
```diff
@@ -18,9 +18,9 @@
 
 public abstract class AbstractPaginatedData<T> implements PaginatedData<T> {
 
-     boolean hasPreviousPage;
-     boolean hasNextPage;
-     Iterable<T> data;
+    Iterable<T> data;
```
yarinvak: should not be private also?

src/main/java/graphql/annotations/connection/SimplePaginatedDataImpl.java
```diff
@@ -0,0 +1,16 @@
+package graphql.annotations.connection;
+
+public class SimplePaginatedDataImpl<T> extends AbstractSimplePaginatedData<T> {
+
+    private long overAll;
```
yarinvak: whats overall means?

src/main/java/graphql/annotations/processor/retrievers/GraphQLFieldRetriever.java
```diff
@@ -175,6 +189,31 @@ private TypeFunction getTypeFunction(Field field, ProcessingElementsContainer co
         return typeFunction;
     }
 
+    private List<GraphQLArgument> getArgumentsForSimpleConnection() {
+        List<GraphQLArgument> args = new ArrayList<>();
```
yarinvak, line 197: In simple pagination we want to receive only index & number of elements to receive

src/main/java/graphql/annotations/processor/retrievers/fieldBuilders/field/FieldDataFetcherBuilder.java
```diff
@@ -33,23 +35,25 @@
...
 
+@SuppressWarnings("ALL")
 public class FieldDataFetcherBuilder implements Builder<DataFetcher> {
     private Field field;
     private DataFetcherConstructor dataFetcherConstructor;
     private GraphQLType outputType;
     private TypeFunction typeFunction;
     private ProcessingElementsContainer container;
-    private boolean isConnection;
+    private TypesConnectionChecker typesConnectionChecker;
```
yarinvak, line 48: why not using a "connectionType" field or something like that? and it will be null if no connection is sent..
Because the TypesConnectionChecker is a little bit confusing

yarinvak (changes requested):
The "SimplePagination"/"SimpleConnection" name is a little bit confusing. 
I would suggest calling it @GraphQLChunk
and "SimpleChunk" instead of connection.

guy120494:
On the way I organized some of the classes, that's why There are a lot of modified classes

src/main/java/graphql/annotations/connection/AbstractPaginatedData.java
guy120494: I don't think so, because the extending class may use the data (not like the pages, where there is no need to use them once you know if they are true or false)

src/main/java/graphql/annotations/connection/SimplePaginatedDataImpl.java
guy120494: the amount of all the entities that can be retrieved

src/main/java/graphql/annotations/processor/retrievers/GraphQLFieldRetriever.java
guy120494: the index is "after" and the number is "first"

guy120494:
@tdraier any thoughts?

src/main/java/graphql/annotations/processor/retrievers/GraphQLFieldRetriever.java
yarinvak: but there is no need in before and last
and after does not mean "starting index" so its not compatible

src/main/java/graphql/annotations/connection/SimplePaginatedDataImpl.java
yarinvak: call it totalCount or overAllCount

src/main/java/graphql/annotations/connection/AbstractPaginatedData.java
yarinvak: so set it to protected

guy120494: done

src/main/java/graphql/annotations/connection/SimplePaginatedDataImpl.java
guy120494: done

tdraier:
I actually have the same kind of needs, but solved it by using a custom relay implementation. The custom relay add some new arguments (instead of first/after, i have limit/offset), the data fetchers still have the @GraphQLConnection with a specific ConnectionFetcher class, that return an extension of Connection with a "nodes" fields (equivalent to your data field). The annotated method returns a object that inherits from PaginatedData (mainly to add the "totalCount", which is not available by default)
The difference is that my connection still have the "default" fields since I inherit from ConnectionFetcher, but you could also remove the "edges" fields and the parameters you don't want by just changing the relay implementation.
Basically it's already possible to customize connections the way you want, without having to change anything in graphql-java-annotation itself. So I'm not sure of the need to add complexity with a new annotation.

guy120494:
That's a good idea, but I don't like the way we change the relay implementation, because what if I want both the default relay for some mutations, and a custom relay for others?
Maybe I should add the relay type in the `@GrahpQLConnection` annotation?

tdraier:
Yes, that's a limitation, currently the same relay implementation is used for all connections. I think it's completely possible to pass the relay implementation to the @GraphQLConnection , and use the one in the container as the default - that would be a nice improvement.

guy120494:
So I'm rejecting this PR, and will open a new one

PR #165, opened 2018-05-02
Fix union

guy120494:
Now union type can have a `TypeResolver`. I also added a default `TypeResolver` that resolves the type by its name, i.e: The DB and API entities must have similar names, up to "API" suffix in the API entity's name

@yarinvak @tdraier as usual, please CR me

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,42 @@ GraphQLInterfaceType object = GraphQLAnnotations.iface(SomeInterface.class);
 An instance of the type resolver will be created from the specified class. If a `getInstance` method is present on the
 class, it will be used instead of the default constructor.
 
+## Defining Unions
+
+To have a union, you must annotate an interface with `@GraphQLUnion`. In the annotation, you must declare all the 
+possible types of the union, and a type resolver.
+If no type resolver is specified, `UnionTypeResovler` is used. It follows this algorithm:
+The resolver assumes the the DB entity's name contains the API entity's name, after removing the "Api"
+suffix from the name of the API entity. If so, it takes the result from the dataFetcher and decides to which
+API entity it should be mapped (according to the name). 
+Example: If you have a `Pet` union type, and the dataFetcher returns `DogDB` (or just `Dog`), the typeResolver
+will check for each API entity if its name (without the "Api" suffix - if the suffix exists) is contained 
+inside the name of the DB entity, so both `Dog` and `DogApi` are valid.
+
+```java
+@GraphQLUnion(possibleTypes={DogApi.class, Cat.class})
+public interface Pet {}
+``` 
+and an example with custom `TypeResovler`:
+```java
+@GraphQLUnion(possibleTypes={DogApi.class, Cat.class}, typeResolver = PetTypeResolver.class)
+public interface Pet {}
+
+
+public class PetTypeResolver implements TypeResolver {
+    @Override
+    GraphQLObjectType getType(TypeResolutionEnvironment env) {
+        Object obj = env.getObject();
+        if(obj instanceof DogDB) {
+            return (GraphQLObjectType) env.getSchema().getType("DogApi");
+        }
+        else {
+            return (GraphQLObjectType) env.getSchema().getType("Cat");
+        }
+      
+    }
+}
+```
 ## Fields
 
 In addition to specifying a field over a Java class field, a field can be defined over a method:
--- a/README.md
+++ b/README.md
@@ -89,6 +89,10 @@ public class PetTypeResolver implements TypeResolver {
     }
 }
 ```
+NOTE: you can have (but not mandatory) a type resolver with constructor that has `Class<?>[]` as the first parameter and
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -54,6 +54,42 @@ GraphQLInterfaceType object = GraphQLAnnotations.iface(SomeInterface.class);
...
+To have a union, you must annotate an interface with `@GraphQLUnion`. In the annotation, you must declare all the 
+possible types of the union, and a type resolver.
+If no type resolver is specified, `UnionTypeResovler` is used. It follows this algorithm:
+The resolver assumes the the DB entity's name contains the API entity's name, after removing the "Api"
+suffix from the name of the API entity. If so, it takes the result from the dataFetcher and decides to which
+API entity it should be mapped (according to the name). 
+Example: If you have a `Pet` union type, and the dataFetcher returns `DogDB` (or just `Dog`), the typeResolver
+will check for each API entity if its name (without the "Api" suffix - if the suffix exists) is contained 
+inside the name of the DB entity, so both `Dog` and `DogApi` are valid.
+
```
tdraier: I completely agree with the fact the possibility to pass a type resolver to a union if needed, but I don't understand the motivation behind changing the previous behaviour of UnionTypeResolver. In my opinion it was quite clear and working for most of the cases. On the other hand in don't understand the example you give in the unit test, where a field of type "Hardware" has a data fetcher which returns an object that is not implementing Hardware. DataFetcher should return a type which is compatible with the field type, and more specifically a type which should match one of the possible types of the union (that the previous UnionTypeResolver was checking). By the way, it would be impossible to return something else with the default MethodDataFetcher.
The issue you are trying to solve (a custom DataFetcher which returns an object different from the field class) could be solved by the TypeResolver you provide, but I don't think it should be the default one. I would rather keep the previous one , working with "isAssignableFrom", and add the new one with a different name ( NameBasedUnionTypeResolver ?)

tdraier (changes requested):
I completely agree with the fact the possibility to pass a type resolver to a union if needed, but I don't understand the motivation behind changing the previous behaviour of UnionTypeResolver. In my opinion it was quite clear and working for most of the cases. On the other hand in don't understand the example you give in the unit test, where a field of type "Hardware" has a data fetcher which returns an object that is not implementing Hardware. DataFetcher should return a type which is compatible with the field type, and more specifically a type which should match one of the possible types of the union (that the previous UnionTypeResolver was checking). By the way, it would be impossible to return something else with the default MethodDataFetcher.
The issue you are trying to solve (a custom DataFetcher which returns an object different from the field class) could be solved by the TypeResolver you provide, but I don't think it should be the default one. I would rather keep the previous one , working with "isAssignableFrom", and add the new one with a different name ( NameBasedUnionTypeResolver ?)

guy120494:
The problem was that `getObject` returns a DB entity, and if you have different classes for DB and API entities (as you should), the type resovler will not match any of the possibilities. The most reasonable defalut behaviour I could think of is trying to match a possibility by name, because in most cases the DB name and API are similar (for example: `CatDB` and `CatApi`)

README.md
guy120494: >On the other hand in don't understand the example you give in the unit test, where a field of type "Hardware" has a data fetcher which returns an object that is not implementing Hardware

Can you say what test? because I dont rememeber doing such a thing

tdraier: Maybe I missed something, but : the Query type defines a `hardwareScreen` field of type `Hardware` - the `getHardwareScreen` method returns null but the field has a custom dataFetcher `ScreenFetcher` (so method won't be called) - the `ScreenFetcher` is actually a `DataFetcher<ScreenDB>`, but `ScreenDB` does not implement `Hardware` (in my opinion it should actually throws an error when detecting this - a field should have a `DataFetcher` of the same type as itself, or compatible, so `DataFetcher<Hardware>` or `DataFetcher<Screen>` ). 
Generally, I don't understand what "DB" objects are for, and the resolution from "ScreenDB" to "Screen" (one possible type of the union) is not really obvious .. ?

tdraier:
Agree that the problem is that getObject returns a "DB" entity. If by DB you mean any internal/backend object, they should not be returned by any of the Api objects - and especially the DataFetcher. DataFetcher should return an object usable by the api.

README.md
yarinvak: I think that guy ment to create a separation between the GraphQL layer and the BL classes. So while Screen is a class returned by the GraphQL Api layer of the service, the datafetcher is able to return an non-api entity (like ScreenDB, which is used inside the BL or is mapped from the database), and this entity is later mapped to the graphql entity by its fields' datafetchers.

I also believe that this DB/API suffixes is a very private case and is not a standard, so we should not add this behaviour as a default one in the UnionTypeResolver.
I support @tdraier and think that UnionTypeResolver should be matched by name/implementation of the interface, and if someone wants a different behaviour he would implement his own UnionTypeResolver.

yarinvak (approved):

Issue #164, opened 2018-04-25
Custom PageInfo


comtihon:
Default PageInfo is
```
#Information about pagination in a connection.
type PageInfo {
  #When paginating forwards, the cursor to continue.
  endCursor: String
  #When paginating forwards, are there more items?
  hasNextPage: Boolean!
  #When paginating backwards, are there more items?
  hasPreviousPage: Boolean!
  #When paginating backwards, the cursor to continue.
  startCursor: String
}
```
F.e. I would like to add additional fields like `TotalPages: Int!`. 
How can I do it?

I tried to:
* implement PaginatedData:
```
@Getter
@AllArgsConstructor
public class CountedPaginatedData<T extends Node> implements PaginatedData<T> {
    private boolean hasPreviousPage;
    private boolean hasNextPage;
    private Long pagesTotal;
    private Iterable<T> data;

    @Override
    public Iterator<T> iterator() { return data.iterator();  }
    @Override
    public boolean hasNextPage() { return hasNextPage; }
    @Override
    public boolean hasPreviousPage() { return hasPreviousPage; }
    @Override
    public String getCursor(T entity) { return entity.id(); }
}
```
* implement PageInfo:
```
@Getter
@RequiredArgsConstructor
public class CountedPageInfo implements PageInfo {
    private final ConnectionCursor startCursor;
    private final ConnectionCursor endCursor;
    private final boolean hasPreviousPage;
    private final boolean hasNextPage;
    private final Long pagesTotal;
}
```
* have my own ConnectionFetcher:
```
public class CountedPaginatedDataConnectionFetcher<T> implements ConnectionFetcher<T> {
...
private PageInfo getPageInfo(List<Edge<T>> edges, PaginatedData<T> paginatedData) {
        if (edges.isEmpty()) {
            return new CountedPageInfo(null, null, false, false, null);
        }
        ConnectionCursor firstCursor = edges.get(0).getCursor();
        ConnectionCursor lastCursor = edges.get(edges.size() - 1).getCursor();
        return new CountedPageInfo(
                firstCursor,
                lastCursor,
                paginatedData.hasPreviousPage(),
                paginatedData.hasNextPage(),
                ((CountedPaginatedData) paginatedData).getPagesTotal()
        );
    }
...
```
* use connection fetcher:
```
@GraphQLField
    @GraphQLConnection(connection = CountedPaginatedDataConnectionFetcher.class)
[...]

PR #166, opened 2018-05-02
Fix interface

guy120494:
I have added a class that is responsible of the "additionalTypes" section in the GraphQL schema.
It has a class named `getAdditionalInterfacesImplementations` that put all the implementations that are not explicitly specified in the schema.
It assumes that you used the `GraphQLAnnotations.object(Query.class)` because it uses the container.
You must first build the schema as usual, and then use the `transform` function to add the additional implementations.
How to use:
```
  GraphQLSchema schema = newSchema().query(GraphQLAnnotations.object(Query.class)).build();
  schema = schema.transform(builder -> builder.additionalTypes(GraphQLAnnotations.additionalTypes(Query.class)));
```
Don't merge yet, as it has not so good performances. What I do here is for every interface I scan the class path for all classes that implement this interface, and it takes a lot of time. Any suggestion of how to make it better is welcomed

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,6 +70,7 @@ gradle.projectsEvaluated {
 dependencies {
     compile 'javax.validation:validation-api:1.1.0.Final'
     compile 'com.graphql-java:graphql-java:7.0'
+    compile 'org.clapper:javautil:3.2.0'
 
     // OSGi
     compileOnly 'org.osgi:org.osgi.core:6.0.0'
--- a/src/main/java/graphql/annotations/processor/GraphQLAnnotations.java
+++ b/src/main/java/graphql/annotations/processor/GraphQLAnnotations.java
@@ -27,8 +27,10 @@ import graphql.annotations.processor.typeFunctions.TypeFunction;
 import graphql.annotations.processor.util.DataFetcherConstructor;
 import graphql.relay.Relay;
 import graphql.schema.GraphQLObjectType;
+import graphql.schema.GraphQLType;
 
 import java.util.Map;
+import java.util.Set;
 
 import static graphql.annotations.processor.util.NamingKit.toGraphqlName;
 
@@ -40,6 +42,7 @@ public class GraphQLAnnotations implements GraphQLAnnotationsProcessor {
 
     private GraphQLObjectHandler graphQLObjectHandler;
     private GraphQLExtensionsHandler graphQLExtensionsHandler;
+    private GraphQLAdditionalTypesHandler graphQLAdditionalTypesHandler;
 
     private ProcessingElementsContainer container;
 
@@ -73,6 +76,13 @@ public class GraphQLAnnotations implements GraphQLAnnotationsProcessor {
         extensionsHandler.setMethodSearchAlgorithm(methodSearchAlgorithm);
         extensionsHandler.setFieldRetriever(fieldRetriever);
 
+        GraphQLAdditionalTypesHandler additionalTypesHandler = new GraphQLAdditionalTypesHandler();
+        additionalTypesHandler.setGraphQLInterfaceRetriever(interfaceRetriever);
+        additionalTypesHandler.setGraphQLObjectInfoRetriever(objectInfoRetriever);
+        additionalTypesHandler.setMethodSearchAlgorithm(methodSearchAlgorithm);
+        additionalTypesHandler.setFieldSearchAlgorithm(fieldSearchAlgorithm);
+
+        this.graphQLAdditionalTypesHandler = additionalTypesHandler;
         this.graphQLObjectHandler = objectHandler;
         this.graphQLExtensionsHandler = extensionsHandler;
         this.container = new ProcessingElementsContainer(defaultTypeFunction);
@@ -104,6 +114,12 @@ public class GraphQLAnnotations implements GraphQLAnnotationsProcessor {
         return instance.graphQLObjectHandler.getObject(object, instance.getContainer());
     }
 
+    public static Set<GraphQLType> additionalTypes(Class<?> root) {
+        GraphQLAnnotations instance = getInstance();
+        return instance.graphQLAdditionalTypesHandler.getAdditionalInterfacesImplementations(root, instance.getContainer());
+
+    }
+
     public void registerTypeExtension(Class<?> objectClass) {
         graphQLExtensionsHandler.registerTypeExtension(objectClass, container);
     }
@@ -116,7 +132,7 @@ public class GraphQLAnnotations implements GraphQLAnnotationsProcessor {
         getInstance().registerType(typeFunction);
     }
 
-    public Map<String, graphql.schema.GraphQLType> getTypeRegistry() {
+    public Map<String, GraphQLType> getTypeRegistry() {
         return container.getTypeRegistry();
     }
 
--- a/src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
+++ b/src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2016 Yurii Rashkovskii
- * <p>
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * <p>
- * http://www.apache.org/licenses/LICENSE-2.0
- * <p>
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
+++ b/src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
@@ -1,3 +1,17 @@
+/**
+ * Copyright 2016 Yurii Rashkovskii
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ */
 package graphql.annotations.processor.retrievers;
 
 import graphql.GraphQLException;
--- a/src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
+++ b/src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
@@ -44,13 +44,11 @@ public class GraphQLAdditionalTypesHandler {
     private GraphQLObjectInfoRetriever graphQLObjectInfoRetriever;
     private SearchAlgorithm methodSearchAlgorithm;
     private SearchAlgorithm fieldSearchAlgorithm;
-    private GraphQLObjectInfoRetriever infoRetriever;
     private ClassFinder classFinder;
 
     public GraphQLAdditionalTypesHandler() {
         classFinder = new ClassFinder();
         classFinder.addClassPath();
-        infoRetriever = new GraphQLObjectInfoRetriever();
     }
 
     public Set<GraphQLType> getAdditionalInterfacesImplementations(Class<?> root, ProcessingElementsContainer container) {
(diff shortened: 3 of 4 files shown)


comtihon:
Ok, as far as I understood, we have `Query` returning `TestIface` interface. While `IfaceFetcher` returns `TestObject`  -  `TestIface`'s implementation.
And query `{ iface { value } }` will return `a`.
In https://github.com/graphql-java/graphql-java-annotations/issues/164 I am trying to access implementation's properties, which are not in the interface.
Consider `TestObject` has `value2` method, which is not in the `TestIface`. Will I be able to call it? (via `value2` or `... on TestObject { value2 }`?
Another question - why is has a performance problem? I thought schema is built only once, on application startup. Or you refer to transform on every transaction?
Why not annotate needed implementation (TestObject, which I am going to use for value2) instead of scanning for all implementations?

guy120494:
As for your first question: yes, you can via `... on TestObject { value2 }` (see `queryForObject_objectImplementsInterfaceButNotExplicitInTheSchema_worksAsExpected` test in `GraphQLInterfaceTest` class).

Your second question: you are right, it is built only once (on server startup), but when you develop you start your application often, and it can be annoying

src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
```diff
@@ -0,0 +1,116 @@
...
+
+@SuppressWarnings("ConstantConditions")
+@Component(service = GraphQLAdditionalTypesHandler.class, immediate = true)
+public class GraphQLAdditionalTypesHandler {
+
+    private GraphQLInterfaceRetriever graphQLInterfaceRetriever;
+    private GraphQLObjectInfoRetriever graphQLObjectInfoRetriever;
+    private SearchAlgorithm methodSearchAlgorithm;
+    private SearchAlgorithm fieldSearchAlgorithm;
+    private GraphQLObjectInfoRetriever infoRetriever;
```
tdraier: Why do we have 2 instances of GraphQLObjectInfoRetriever here ? graphQLObjectInfoRetriever should be enough

src/main/java/graphql/annotations/processor/GraphQLAnnotations.java
```diff
@@ -73,6 +76,13 @@ public GraphQLAnnotations() {
         extensionsHandler.setMethodSearchAlgorithm(methodSearchAlgorithm);
         extensionsHandler.setFieldRetriever(fieldRetriever);
 
+        GraphQLAdditionalTypesHandler additionalTypesHandler = new GraphQLAdditionalTypesHandler();
+        additionalTypesHandler.setGraphQLInterfaceRetriever(interfaceRetriever);
+        additionalTypesHandler.setGraphQLObjectInfoRetriever(objectInfoRetriever);
+        additionalTypesHandler.setMethodSearchAlgorithm(methodSearchAlgorithm);
+        additionalTypesHandler.setFieldSearchAlgorithm(fieldSearchAlgorithm);
+
```
tdraier, line 84: Can you also add the GraphQLAdditionalTypesHandler to GraphQLAnnotationsComponent ? A simple field with getter/setter/unsetter as the others fields should be ok

src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
```diff
@@ -0,0 +1,116 @@
...
+                    }
+                } catch (ClassNotFoundException e) {
+                    throw new GraphQLException(e);
+                }
+            }
+        }
+        return additionalFields;
+    }
+
+    public void setGraphQLObjectInfoRetriever(GraphQLObjectInfoRetriever graphQLObjectInfoRetriever) {
```
tdraier, line 99: The @Reference annotations and the unsetter will be needed to correctly initialize the osgi component

src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
```diff
@@ -0,0 +1,116 @@
...
+        classFinder = new ClassFinder();
+        classFinder.addClassPath();
+        infoRetriever = new GraphQLObjectInfoRetriever();
+    }
+
+    public Set<GraphQLType> getAdditionalInterfacesImplementations(Class<?> root, ProcessingElementsContainer container) {
+        Set<GraphQLType> additionalFields = new HashSet<>();
+
+        additionalFields.addAll(getNewImplementations(container, classFinder, root));
+        for (Method method : graphQLObjectInfoRetriever.getOrderedMethods(root)) {
```
tdraier, line 58: In my opinion we should not go through the different fields but rather let the user explicitely pass the interfaces he wants to parse. I don't really see the benefit and it could be confusing and error-prone, as you don't necessarily want to consider all fields as interfaces with automatically discovered implementations.

build.gradle
```diff
@@ -70,6 +70,7 @@ gradle.projectsEvaluated {
 dependencies {
     compile 'javax.validation:validation-api:1.1.0.Final'
     compile 'com.graphql-java:graphql-java:7.0'
+    compile 'org.clapper:javautil:3.2.0'
```
tdraier: This dependency is not a valid osgi bundle and is not correctly parsed by the osgi plugin, this can be an when deploying in a container. At least it would be good to have the dependency optional for osgi, but I don't really know how to do that with gradle

src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
guy120494: Will fix it

src/main/java/graphql/annotations/processor/GraphQLAnnotations.java
guy120494: Can you tell me exactly what to do? I don't know anything in OSGI

src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
guy120494: Can you tell me exactly what to do? I don't know anything in OSGI

src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
guy120494: The problem is the following:
We build the schema like a tree, starting from query and go to its field and so on. If there is an implementation of some interface that is not in the schema (only the interface is in the schema), I will not get to that implementation class, so even with your solution the problem will still exists. The only way I can think of finding the classs is to search the class path for all the classes that implement the interface. For more details about the problem, take a look [here](https://github.com/graphql-java/graphql-java/issues/950)

build.gradle
guy120494: So what can I do? I need this dependency to be able to search the class path

src/main/java/graphql/annotations/processor/GraphQLAnnotations.java
tdraier: I can do a PR once it will be merged

src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
tdraier: Ok, I understand. Do you think it would make sense to have an annotations on the interface that could be tested when calling getNewImplementations ? Or an attribute in the GraphQLTypeResolver ( something like automaticallyDiscoverSubClasses ) ?
And doing this, it could also be safe to automatically call graphQLAdditionalTypesHandler.getAdditionalInterfacesImplementations everytime we register an object class, so that you don't need to explicitely call it. What do you think ?

build.gradle
tdraier: Not sure, I can have a look at it after your merge. I will have to do some tests in an OSGi environment.

src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
guy120494: I don't understand what you mean. Do you want to have an attribute in `@GraphQLTypeResolver` that says to search for implementation, So if it is off we don't do it? 
Well, I think I can add this attribute, but I also think that in most cases people will set this attribute to "true", because if you have an interface, its implementations will not be explicitly in the schema. But I don't mind adding that

src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
guy120494: Done

src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
tdraier: Until now I was adding the implementation manually - so at least for compatibility, or if we explicitely do not want to parse all classes, the attribute could be useful. Default value can be to do the parsing, no problem with that, in most cases it should really help.
What about calling it automatically everytime we create a new type ?

src/main/java/graphql/annotations/processor/GraphQLAnnotations.java
```diff
@@ -73,6 +76,13 @@ public GraphQLAnnotations() {
         extensionsHandler.setMethodSearchAlgorithm(methodSearchAlgorithm);
         extensionsHandler.setFieldRetriever(fieldRetriever);
 
+        GraphQLAdditionalTypesHandler additionalTypesHandler = new GraphQLAdditionalTypesHandler();
```
yarinvak, line 79: Maybe create a constructor that gets the parameters below? So we wont need these five lines..

src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
```diff
@@ -0,0 +1,114 @@
...
+                additionalFields.addAll(getAdditionalInterfacesImplementations(field.getType(), container));
+            }
+        }
+
+        return additionalFields;
+    }
+
+    private Set<GraphQLType> getNewImplementations(ProcessingElementsContainer container, ClassFinder classFinder, Class<?> aClass) {
+        Set<GraphQLType> additionalFields = new HashSet<>();
+        if (container.getTypeRegistry().containsKey(graphQLObjectInfoRetriever.getTypeName(aClass)) && aClass.isInterface() && aClass.isAnnotationPresent(GraphQLTypeResolver.class)) {
```
yarinvak, line 81: Check the last two conditions at first

src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
```diff
@@ -0,0 +1,114 @@
...
+            ClassFilter classFilter = new SubclassClassFilter(aClass);
+            List<ClassInfo> foundClasses = new ArrayList<>();
+            classFinder.findClasses(foundClasses, classFilter);
+            for (ClassInfo classInfo : foundClasses) {
+                try {
+                    if (!(container.getTypeRegistry().containsKey(graphQLObjectInfoRetriever.getTypeName(Class.forName(classInfo.getClassName()))))) {
```
yarinvak, line 87: This condition is too long and not readable

src/main/java/graphql/annotations/processor/retrievers/GraphQLAdditionalTypesHandler.java
```diff
@@ -0,0 +1,114 @@
...
+
+    public GraphQLAdditionalTypesHandler() {
+        classFinder = new ClassFinder();
+        classFinder.addClassPath();
+    }
+
+    public Set<GraphQLType> getAdditionalInterfacesImplementations(Class<?> root, ProcessingElementsContainer container) {
+        Set<GraphQLType> additionalFields = new HashSet<>();
+
+        additionalFields.addAll(getNewImplementations(container, classFinder, root));
```
yarinvak, line 57: Example when implementations of root class is needed?

yarinvak:
I believe that its not the annotations library's job to find for implementations and that the implementations should be registered manually by the implementer

guy120494:
@tdraier , what do you think? I think it is a nice addition to have the possibility to register all the implementations automatically (because as I see it, implementations are part of the object and part of the schema), @yarinvak claims that it is not part of this library, because the library's job is to create the `GraphQLType`, and then the user register the `Query` and `Mutation` by itself, so if he wants an implementation of an interface, he should create the associated `GraphQLType` and then put it in `additionalTypes` - but again, he must do that to all the implementations (and there may be a lot)

build.gradle
danieltaub96: hi, look at this one :[reflections]( https://github.com/ronmamo/reflections)

yarinvak:
If the user has a lot of implementations, he can do the classpath search by himself, and create a list of types to pass to the additionalTypes in the schema creation. But as a user i do not expect that when I create a GraphQLObject out of a java class, that the library searches in my classpath for implementations. Because as long as I neither referenced the interface's implementation in my java class nor passed the implementation to the additionalTypes, I don't expect anyone to be familiar with my implementations.
And even if you say it's an optional feature - it's a nice util but I don't think it belongs here

PR #167, opened 2018-05-07
DataFetcherConstructor args constructor as the annotation + tests

danieltaub96:
@GraphQLDataFetcher have args (String[]) fields, and the class thats implements DataFetcher should expect to get those args with String[] constructor...
of course that it's have Backwards compatibility
wrote some tests for the DataFetcherConstructor class

### DIFF
--- a/src/main/java/graphql/annotations/processor/util/DataFetcherConstructor.java
+++ b/src/main/java/graphql/annotations/processor/util/DataFetcherConstructor.java
@@ -39,9 +39,9 @@ public class DataFetcherConstructor {
             return newInstance(annotatedDataFetcher.value());
         } else {
             try {
-                final Constructor<? extends DataFetcher> ctr = annotatedDataFetcher.value().getDeclaredConstructor(
-                        stream(args).map(v -> String.class).toArray(Class[]::new));
-                return constructNewInstance(ctr, (Object[]) args);
+                // Get the String[] constructor
+                final Constructor<? extends DataFetcher> ctr = annotatedDataFetcher.value().getDeclaredConstructor(String[].class);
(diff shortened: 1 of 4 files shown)


src/main/java/graphql/annotations/processor/util/DataFetcherConstructor.java
```diff
@@ -38,14 +38,31 @@ public DataFetcher constructDataFetcher(String fieldName, GraphQLDataFetcher ann
...
+                    // filter only constructor that have the same args as the annotation args
+                    // or that get String[], String... args
+                    .filter(x -> (x.getParameterCount() == args.length) ||
+                            (x.getParameterTypes().length == 1 && x.getParameterTypes()[0] == String[].class))
+                    .map(x -> {
+                        try {
+                            Constructor<? extends DataFetcher> constructor;
+                            if (x.getParameterTypes().length == 1 && x.getParameterTypes()[0] == String[].class) {
+                                constructor = annotatedDataFetcher.value().getDeclaredConstructor(String[].class);
+                                return constructNewInstance(constructor, new Object[]{args});
```
guy120494, line 51: new Object[]{args} -> (Object[]) args

danieltaub96: it's in purpose that way 
new Object[]{args} create a single variable of object[],
and (Object[]) is a casting that casts to a object array, and the String[] ctor expecting only one argument...

yarinvak:
Please udpate the readme with all the possible options of how the constructor of datafetche can look like

danieltaub96:
done

PR #171, opened 2018-05-10
Method name is not changed when it becomes field

guy120494:
The name is not changed when the method becomes field
i.e the name of the field that is generated from a method is exactly the same name of the method
(We don't remove the get|set|is prefix)


yarinvak:
I believe this pr will solve a lot of confusion. When someone wants to alter names he should use ``@GraphQLName``

Issue #174, opened 2018-05-24
Bug in Methoddatafetcher


yarinvak:
In method data fetcher we have this line:
```java
Object value = getFieldValue(environment.getSource(), method.getName());
```
so even if we have altered the name of the method, with ``@GraphQLName`` or ``@GraphQLPrettify``, we do not get the field by this name but by the method name..

PR #175, opened 2018-06-08
Fix method data fetcher

yarinvak:
fix https://github.com/graphql-java/graphql-java-annotations/issues/174

### DIFF
--- a/src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
+++ b/src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2016 Yurii Rashkovskii
- * <p>
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * <p>
- * http://www.apache.org/licenses/LICENSE-2.0
- * <p>
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
+++ b/src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- * http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
```diff
@@ -135,9 +136,51 @@ private Object buildArg(Type p, GraphQLType graphQLType, Object arg) {
         }
     }
 
-    private Object getFieldValue(Object source, String fieldName) throws IllegalAccessException, NoSuchFieldException {
-        Field field = source.getClass().getDeclaredField(fieldName);
-        field.setAccessible(true);
-        return field.get(source);
+    private Object getGraphQLFieldValue(Object source, String fieldName) throws IllegalAccessException, NoSuchFieldException, InvocationTargetException {
```
guy120494, line 139: Separate to few methods, one that deals with fields and one that deals with methods

src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
```diff
@@ -135,9 +136,51 @@ private Object buildArg(Type p, GraphQLType graphQLType, Object arg) {
...
+        Method method = getMethod(source.getClass(), fieldName, "");
+        Method getMethod = getMethod(source.getClass(), fieldName, "get");
```
guy120494: call this method only if necessary (reflection takes a lot of time)

src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
```diff
@@ -135,9 +136,51 @@ private Object buildArg(Type p, GraphQLType graphQLType, Object arg) {
...
+        Method isMethod = getMethod(source.getClass(), fieldName, "is");
```
guy120494: call this method only if necessary (reflection takes a lot of time)

src/main/java/graphql/annotations/processor/util/PrefixesUtil.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ */
+package graphql.annotations.processor.util;
+
+public class PrefixesUtil {
+    public static String createPrefix(String prefix, String propertyName) {
```
guy120494: The name of the function doesn't explain what the function does

src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
```diff
@@ -135,9 +136,51 @@ private Object buildArg(Type p, GraphQLType graphQLType, Object arg) {
...
+        }
+
+        return method;
     }
+
+    private Field getField(Class<?> clazz, String name) {
+        Field field = null;
+        while (clazz != null && field == null) {
+            try {
+                field = clazz.getDeclaredField(name);
```
guy120494, line 186: should it be "getDeclaredField" or "getField"?

guy120494 (changes requested):

src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
yarinvak: getField returns only public fields (but for entire hirearchy), but we need private ones as well. So this method runs over the entire hirearchy and searches for private methods with this name.

PR #184, opened 2018-07-31
Make it possible to provide custom prefixes AND postfixes for input objects

sergehuber:
Make it possible to provide custom prefixes AND postfixes for input object name generation. This is setup on the container.

The defaults were not changed and are equivalent to :
container.setInputPrefix("Input");
container.setInputPostfix("";

If you want to change the naming generation to something more "compliant" with GraphQL's naming conventions you can change the configuration to:

container.setInputPrefix("");
container.setInputPostfix("Input");

So for example, if generating a name for the input type generated from the class Books it will be called InputBooks in the first configuration and BooksInput in the second configuration. This is also very useful if you are using prefixes to namespace your classes. For example, ACMEBooks will be InputACMEBooks in the first configuration and ACMEBooksInput in the second configuration.

As you can see the prefix and postfix configuration is also very flexible and the "Input" word is just a configuration variable now.

By default, however, nothing changes so existing projects should not break.

Some unit tests were also added to make sure that the generation works properly.

Please don't hesitate if you have any questions or comments about this PR, I will do my best to answer swiftly.

### DIFF
--- a/src/main/java/graphql/annotations/processor/ProcessingElementsContainer.java
+++ b/src/main/java/graphql/annotations/processor/ProcessingElementsContainer.java
@@ -27,7 +27,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.Stack;
 
-import static graphql.annotations.processor.util.InputPropertiesUtil.DEFAULT_INPUT_POSTFIX;
+import static graphql.annotations.processor.util.InputPropertiesUtil.DEFAULT_INPUT_SUFFIX;
 import static graphql.annotations.processor.util.InputPropertiesUtil.DEFAULT_INPUT_PREFIX;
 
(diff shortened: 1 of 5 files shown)


yarinvak:
It looks very nice, thank you.
@guy120494 can you look at this also?

guy120494:
Looks nice.
Only one request: please change "postfix" to "suffix" (I know it's a petty request :) )

sergehuber:
Actually that change does make sense, I'll do it now.

guy120494:
Please resolve the conflict

sergehuber:
I've just resolved the conflict, I will do the change from postfix to suffix next.

sergehuber:
Ok I've renamed postfix to suffix. I hope you can accept this PR now :)

PR #192, opened 2018-11-17
Graphql java 11

yarinvak:


sergehuber:
Looks ok to me. Thanks. 
What was the problem ?

PR #204, opened 2019-04-05
Allow explicit GraphQLInputObjectType creation from annotated classes

cverdoes:
In some instances, when manually defining fields with input arguments there's a need for some way to explicitly create GraphQLInputObjectTypes from complex objects using annotations much like can be done for GraphQLObjectTypes. This PR exposes such a method in the GraphQLAnnotations API.


yarinvak:
Why is this needed? GraphQLInputObject type are generated if you use your type as an argument of some graphql field automatically

cverdoes:
Maybe I'm missing something. Consider the test case I added - how would you achieve that using fields with input arguments?

cverdoes:
Ok, I've now reviewed the code a bit more closely, and realized I've been using it wrong :)

PR #214, opened 2019-06-20
Remove static method to use the current instance

javiergamarra:
I think* that this doesn't need to be static and, right now, is prohibiting overrides of the method for classes that extend GraphQLObjectInfoRetriever.

Thank you for your library, nice work :)


javiergamarra:
Thank you! :)

PR #250, opened 2020-03-27
fix @GraphQLConnection for [ObjectType!]!

asinbow:
For easing the frontend development in strict typed system like `typescript` or `flowjs`, we usually use `double @GraphQLNonNull like [ObjectType!]!`. But there is a small bug in `ConnectionUtil.java` that it doesn't unwrap the internal NonNull`!`. This PR is for it , thanks for reviewing.

### DIFF
--- a/src/main/java/graphql/annotations/processor/util/ConnectionUtil.java
+++ b/src/main/java/graphql/annotations/processor/util/ConnectionUtil.java
@@ -33,13 +33,24 @@ public class ConnectionUtil {
     private static final List<Class> TYPES_FOR_CONNECTION = Arrays.asList(GraphQLObjectType.class, GraphQLInterfaceType.class, GraphQLUnionType.class, GraphQLTypeReference.class);
 
     public static boolean isConnection(AccessibleObject obj, GraphQLType type) {
+        if (!obj.isAnnotationPresent(GraphQLConnection.class)) {
+          return false;
+        }
+
         if (type instanceof graphql.schema.GraphQLNonNull) {
             type =  ((GraphQLNonNull) type).getWrappedType();
         }
-        final GraphQLType actualType = type;
-        boolean isValidGraphQLTypeForConnection = obj.isAnnotationPresent(GraphQLConnection.class) &&
-                actualType instanceof GraphQLList && TYPES_FOR_CONNECTION.stream().anyMatch(aClass ->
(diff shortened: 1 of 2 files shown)


asinbow:
~~Totally new about this repository.
I want to improve unit test for it, any suggestion?~~

Added my tests for it. I think it's ready for reviewing.

src/main/java/graphql/annotations/processor/util/ConnectionUtil.java
```diff
@@ -33,13 +33,24 @@
...
+
+        if (type instanceof graphql.schema.GraphQLNonNull) {
+            type =  ((GraphQLNonNull) type).getWrappedType();
+        }
+        if (!(type instanceof GraphQLList)) {
+          return false;
+        }
+
+        type = ((GraphQLList) type).getWrappedType();
         if (type instanceof graphql.schema.GraphQLNonNull) {
```
asinbow, line 48: An extra unwrap for types like `[ObjectType!]!`.

src/test/java/graphql/annotations/connection/GraphQLConnectionTest.java
```diff
@@ -201,6 +207,19 @@ public void methodNonNull() {
         testResult("getNonNullObjs", result);
     }
 
+    @Test
+    public void methodDoubleNonNull() {
```
asinbow, line 211: test double `@GraphQLNonNull` like `[ObjectType!]!`.

src/test/java/graphql/annotations/connection/GraphQLConnectionTest.java
```diff
@@ -396,4 +422,42 @@ public String getCursor(Obj entity) {
 
     }
 
+    /*
+     * double @GraphQLNonNull on @GraphQLList like [ObjectType!]!
+     */
+    public static class DoubleNonNullIterableTypeFunction implements TypeFunction {
```
asinbow, line 428: As I have experimented, this is the best way to implement `double @GraphQLNonNull like [ObjectType!]!`.
How about to make it builtin?

yarinvak: interesting.. we have to think about it (I don't think we handle double graphql non-null at all, not only in connection)

yarinvak (approved):

PR #258, opened 2020-09-15
Gradle update

BlasiusSecundus:
- update to latest Gradle (6.6.1)
- add the official Gradle Wrapper Validator action (https://github.com/gradle/wrapper-validation-action)
- set distribution type to "ALL" (instead of the default bin)


Fgerthoffert:
@BlasiusSecundus We're helping with maintenance for the project, could you tell us a bit more about the reasoning behind the change (other than being on a more recent version).

This PR is also 2 years old, would you consider/need updating to more recent version of graddle.

Without an answer within the next 2 months, we're going to close this PR.

Thanks,

BlasiusSecundus:
Sure, I can update the MR, if there is interest to merge it.

As for the reasons:
- regarding the newer gradle version, there was no specific reason, other than it is generally good practice to keep things up to date
- Gradle Wrapper Validator is a security measure, see the linked description
- as for ALL (vs. the default BIN) distribution - ALL includes the sources and documentation too, and this may help certain IDEs to better support Gradle build scripts; the slight drawback is that the download is obviously somewhat larger than the bin version

sergehuber:
Ok your proposal sounds great. If you can update the PR we will review it asap.

Regards,
  Serge

BlasiusSecundus:
Sure, I'll update it. It may take some time as I probably have to adjust the build script too. (Gradle is now at version 8.1.1, so probably there are some deprecation and breaking changes in between.)

Fgerthoffert:
Hi @BlasiusSecundus,

Did you get a chance to work on the PR ?

BlasiusSecundus:
Not yet really, but it is on my backlog. In the meanwhile, I created a separate PR (https://github.com/Enigmatis/graphql-java-annotations/pull/293) just with the GitHub workflow update.

sergehuber:
FYI, We have since updated the Gradle version to 8.3 in the master branch so you'll want to update your PR to reflect this. The wrapper was also updated but the validation action was for example not added.

sergehuber:
I see we have the validation both here and in the other PR. Maybe we could just keep the separate PR for validation and remove it from this one ?

BlasiusSecundus:
> FYI, We have since updated the Gradle version to 8.3 in the master branch so you'll want to update your PR to reflect this. The wrapper was also updated but the validation action was for example not added.

Considering that I think this MR is now obsolete and can be closed. Meanwhile I have rebased the other one with the validator.

sergehuber:
OK thanks @BlasiusSecundus

PR #272, opened 2021-08-04
(deps) Updating to latest graphql-java:17.1 library

setchy:
Updating to the recently released [graphql-java:17.0 library](https://github.com/graphql-java/graphql-java/releases/tag/v17.0)

As part of 17.0, the following breaking changes occured
* The deprecated ` BatchedExecutionStrategy` was removed.  As such, I've removed `@GraphQLBatched`
* Non-standard scalars (BigInteger, BigDecimal, Char, Byte, Long, Short) were all removed.   This was communicated within the [15.0 release notes](https://github.com/graphql-java/graphql-java/releases/tag/v15.0) and was finalized in 17.0.  As such, I have removed the same in this PR

Appreciate any feedback

PR #277, opened 2021-08-20
Feature/graphql java 17.1

setchy:
#272 but with passing tests


setchy:
@yrashk - should be good to go now

PR #279, opened 2021-08-28
Adopt is dead, long live Zulu!

geertjanw:


yarinvak:
would appreciate elaboration on your change

geertjanw:
Adopt no longer exists. Two options for v2 right now are 'temurin' or 'zulu'. The first one, 'temurin' is new, and doesn't have all the releases, e.g., the Temurin history starts from 8.0.302, 11.0.12, and 16.0.2, while Zulu supports all versions of these releases.

geertjanw:
Right now you're testing only on '8', which is not a fixed version, but that is whatever the latest version of '8' is. You're recommended as a best practise to set the latest version of a release, as well as a fixed version, so that when your build breaks and the fixed version is green while the latest version is red, you'll know that the broken build is because of the latest version of the JDK and not because of anything in your own code.

Here's an example:

https://github.com/HdrHistogram/HdrHistogram/blob/master/.github/workflows/maven.yml

.github/workflows/build.yml
```diff
@@ -14,7 +14,7 @@ jobs:
         uses: actions/setup-java@v2
         with:
           java-version: '8'
-          distribution: 'adopt'
+          distribution: 'zulu'
```
Fgerthoffert, line 17: Could you also update to zulu in the publish workflow ?

Fgerthoffert (changes requested):
We're helping on project maintenance.

Once the change is done we will merge your PR. Without answer within 2 months, we will close this PR and create a separate one for the update to zulu.

PR #288, opened 2022-10-12
Improving GraphQL type handling and upgrading testng

manikandan-ravikumar:
# Partial update to entities
The purpose of this PR is to enable updation of selective entity fields. By introducing `Optional`, it also differentiates `null` from `undefined`

Let's assume an entity with 5 fields and the requirement is to update only one field to `null`. We can set the value of that field to `null` and omit other fields in a mutation query.

# Changelog
- Fixing import issues with `javax.xml.bind.DatatypeConverter`
- Bump testng version from `6.9.10` to `7.4.0`
- New `GraphQLUndefined` type
- Following improvements in `MethodDataFetcher`:
    1. Improved `Nullable` handling using `Optional`
    2. Handle `Optional` field types
    3. Support `RawType` in additional to `WrpperType`
- `GraphQLInputTest` updated to test `Optional` field types in GraphQL queries

### DIFF
--- a/src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
+++ b/src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
@@ -12,24 +12,38 @@
  */
 package graphql.annotations.dataFetchers;
 
-import graphql.annotations.annotationTypes.GraphQLConstructor;
-import graphql.annotations.annotationTypes.GraphQLInvokeDetached;
-import graphql.annotations.annotationTypes.GraphQLName;
-import graphql.annotations.processor.ProcessingElementsContainer;
-import graphql.annotations.processor.typeFunctions.TypeFunction;
-import graphql.schema.*;
-
-import java.lang.reflect.*;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
-
 import static graphql.annotations.processor.util.NamingKit.toGraphqlName;
 import static graphql.annotations.processor.util.PrefixesUtil.addPrefixToPropertyName;
 import static graphql.annotations.processor.util.PrefixesUtil.extractPrefixedName;
 import static graphql.annotations.processor.util.ReflectionKit.constructNewInstance;
 import static graphql.annotations.processor.util.ReflectionKit.newInstance;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.lang.reflect.Parameter;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import graphql.annotations.annotationTypes.GraphQLConstructor;
+import graphql.annotations.annotationTypes.GraphQLInvokeDetached;
+import graphql.annotations.annotationTypes.GraphQLName;
+import graphql.annotations.processor.ProcessingElementsContainer;
+import graphql.annotations.processor.typeFunctions.TypeFunction;
+import graphql.annotations.type.GraphQLUndefined;
+import graphql.schema.DataFetcher;
+import graphql.schema.DataFetchingEnvironment;
+import graphql.schema.GraphQLInputObjectType;
+import graphql.schema.GraphQLList;
+import graphql.schema.GraphQLType;
 
 
 /**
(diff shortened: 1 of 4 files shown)


manikandan-ravikumar:
Hi @yrashk or anyone,
Is it possible to push this PR further. We are planning to get rid of the fork that we have been using in our project and directly include this repo as a dependency. I have another related PR here https://github.com/Enigmatis/graphql-java-annotations/pull/287. I would really appreciate it if anyone could take a look 🙏

src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
```diff
@@ -118,40 +132,67 @@ private Object[] invocationArgs(DataFetchingEnvironment environment, ProcessingE
         return result.toArray();
     }
 
+    @SuppressWarnings("ConstantConditions")
     private Object buildArg(Type p, GraphQLType graphQLType, Object arg) {
-        if (arg == null) {
+        Optional<Object> optionalArg = Optional.ofNullable(arg);
+        if (!optionalArg.isPresent()) {
             return null;
         }
```
tdraier, line 140: This creation of optional is useless and makes the code much more confusing and complex to understand. You just test that optional isPresent, otherwise return null. An optional which always have a value is not an optional anymore.

src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
```diff
@@ -118,40 +132,67 @@ private Object[] invocationArgs(DataFetchingEnvironment environment, ProcessingE
...
+                    list.add(buildArg(subType, wrappedType, item));
+                }
+                return list;
+            }
+        } else if (p instanceof ParameterizedType && ((ParameterizedType) p).getRawType() == Optional.class) {
+            Type subType = ((ParameterizedType) p).getActualTypeArguments()[0];
+            if (arg == null) {
+                return null;
+            } else {
+                return Optional.ofNullable(buildArg(subType, new GraphQLUndefined(), arg));
```
tdraier, line 192: Here I think we have a major issue : the second parameter should be graphqlType, we are just creating an optional of the arg parameter. If subtype is another parameterized type (a list), it won't be correctly parsed. Anyway it's strange to use GraphqlUndefined here since the value is defined.

src/main/java/graphql/annotations/type/GraphQLUndefined.java
```diff
@@ -0,0 +1,49 @@
...
+import graphql.schema.GraphQLSchemaElement;
+import graphql.schema.GraphQLType;
+import graphql.schema.GraphQLTypeVisitor;
+import graphql.schema.SchemaElementChildrenContainer;
+import graphql.util.TraversalControl;
+import graphql.util.TraverserContext;
+
+import java.util.List;
+
+public class GraphQLUndefined implements GraphQLType {
```
tdraier, line 24: I don't understand the meaning or usage of this class. It's only used in buildArg, in a value that will be thrown away just after.

src/test/java/graphql/annotations/GraphQLInputTest.java
```diff
@@ -191,6 +250,19 @@ public void query() {
...
+
+        GraphQL graphQL = GraphQL.newGraphQL(schema).build();
+        ExecutionResult result = graphQL.execute("{ something(code: {firstField:\"a\",secondField:\"b\"}) otherthing(code: {secondField:\"c\"}) listthings somethingWithOneField(code: {}) }", new QueryUndefinedParameter());
+        assertTrue(result.getErrors().isEmpty());
+        assertEquals(((Map<String, String>) result.getData()).get("something"), "ab");
+        assertEquals(((Map<String, String>) result.getData()).get("otherthing"), "c");
+        assertEquals(((Map<String, String>) result.getData()).get("listthings"), "was null");
+        assertEquals(((Map<String, String>) result.getData()).get("somethingWithOneField"), "was undefined");
+    }
+
```
tdraier, line 265: Since the goal of this story (from my understanding) is to make the distinction between {} and {value: null} , we should at least have a test that shows it. The use cases described here already work without the proposed changes.

src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
```diff
@@ -118,40 +132,67 @@ private Object[] invocationArgs(DataFetchingEnvironment environment, ProcessingE
...
             }
-
-            return list;
         } else {
-            return arg;
+            return optionalArg.orElse(null);
```
tdraier, line 195: again, `return arg` was much more clear

src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
```diff
@@ -118,40 +132,67 @@ private Object[] invocationArgs(DataFetchingEnvironment environment, ProcessingE
...
             Parameter[] parameters = constructor.getParameters();
-            if (parameters.length == 1 && parameters[0].getType().isAssignableFrom(arg.getClass())) {
-                return constructNewInstance(constructor, arg);
+
+            if (parameters.length == 1 && optionalArg.isPresent() && parameters[0].getType().isAssignableFrom(optionalArg.get().getClass())) {
+                if (parameters[0].getType().isAssignableFrom(Optional.class)) {
+                    return constructNewInstance(constructor, arg);
+                } else {
+                    return constructNewInstance(constructor, optionalArg.orElse(null));
+                }
```
tdraier, line 155: I don't see how this case can work. The constructor has one single parameter which is assignable from `arg.getClass()` (Map, for example) and at the same time Optional.class , and in that case , you don't pass the optional but the arg ? otherwise, you keep the previous behaviour , equivalent to `return constructNewInstance(constructor, arg);` ? In that case, rather keep the previous code

src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
```diff
@@ -118,40 +132,67 @@ private Object[] invocationArgs(DataFetchingEnvironment environment, ProcessingE
...
             } else {
                 List<Object> objects = new ArrayList<>();
-                Map map = (Map) arg;
+                Map map = (Map) optionalArg.orElseGet(Collections::emptyMap);
```
tdraier, line 158: OptionalArg isPresent (you test that at the beginning). Keep `Map map = (Map) arg;`

src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
```diff
@@ -118,40 +132,67 @@ private Object[] invocationArgs(DataFetchingEnvironment environment, ProcessingE
...
+                Type subType = ((ParameterizedType) p).getActualTypeArguments()[0];
+                GraphQLType wrappedType = ((GraphQLList) graphQLType).getWrappedType();
 
-            for (Object item : ((List) arg)) {
-                list.add(buildArg(subType, wrappedType, item));
+                for (Object item : ((List) optionalArg.orElseGet(Collections::emptyList))) {
+                    list.add(buildArg(subType, wrappedType, item));
+                }
+                return list;
+            }
```
tdraier, line 186: Keep this whole case unchanged, and move the next case `(p instanceof ParameterizedType && ((ParameterizedType) p).getRawType() == Optional.class)` before the list one

tdraier (changes requested):
Hi, thanks for the PR !

We're now helping on project maintenance, I have reviewed and tested your code. I was quite confused trying to understand what you were trying to achieve, the test case does not reflect any new possibility you introduced. 
In the end,  I think the support for Optional can be achieved by just adding a new case in buildArg. 
Also we did not understand what was the GraphqlUndefined - from what I've seen the class itself is actually never really used .. ?
See comments inline

tdraier (changes requested):
Hi, thanks for the PR !

We're now helping on project maintenance, I have reviewed and tested your code. I was quite confused trying to understand what you were trying to achieve, the test case does not reflect any new possibility you introduced. 
In the end,  I think the support for Optional can be achieved by just adding a new case in buildArg. 
Also we did not understand what was the GraphqlUndefined - from what I've seen the class itself is actually never really used .. ?
See comments inline

Fgerthoffert:
Hi @manikandan-ravikumar,

We will be closing the PR for the time being, there are interesting improvements in your initial proposal but it required some work to be integrated into the codebase.

We don't have the bandwidth to do these changes ourselves for now, but if you're still interested, feel free to re-open the PR and submit the requested changes.

Thanks,

PR #289, opened 2023-01-13
Let ID be every type, since it is allowed by

EugenMayer:
According to https://graphql.org/learn/schema/#scalar-types the ID type can be any type that can be represented by a string, not just a int/Integer/String. For example a Long is a common scalar that can be used for an ID and is perfectly allowed.


sergehuber:
Thanks for the PR, we have merged it.

PR #301, opened 2024-03-20
Use actual graphql name on extending type

geichelberger:
If a type is extended via a `GraphQLTypeExtension` annotation, the simple class name is used instead of checking whether a GraphQLName annotation is available; if the GraphQL name differs from the class name, the field cannot be fetched.

Example:

All the fields from the `GqlCurrentUser` class will be added with `CurrentUser` as the parent name. While handling the extension, the fields will be registered with `GqlCurrentUser` as the parent name, which does not match the actual parent.

```java
@GraphQLName(GqlCurrentUser.TYPE_NAME)
public class GqlCurrentUser {

  public static final String TYPE_NAME = "CurrentUser";

}

@GraphQLTypeExtension(GqlCurrentUser.class)
public class EventCurrentUserExtension {
}
```

https://github.com/Enigmatis/graphql-java-annotations/blob/74dd8c8e4da73b98927d988e4e62b12a733d6fd3/src/main/java/graphql/annotations/processor/typeBuilders/OutputObjectBuilder.java#L82

vs 

https://github.com/Enigmatis/graphql-java-annotations/blob/74dd8c8e4da73b98927d988e4e62b12a733d6fd3/src/main/java/graphql/annotations/processor/retrievers/GraphQLExtensionsHandler.java#L62

### DIFF
--- a/src/main/java/graphql/annotations/processor/retrievers/GraphQLExtensionsHandler.java
+++ b/src/main/java/graphql/annotations/processor/retrievers/GraphQLExtensionsHandler.java
@@ -44,6 +44,7 @@ public class GraphQLExtensionsHandler {
 
     public List<GraphQLFieldDefinition> getExtensionFields(Class<?> object, List<String> definedFields, ProcessingElementsContainer container) throws CannotCastMemberException {
         List<GraphQLFieldDefinition> fields = new ArrayList<>();
+        String typeName = graphQLObjectInfoRetriever.getTypeName(object);
         if (container.getExtensionsTypeRegistry().containsKey(object)) {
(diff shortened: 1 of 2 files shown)


geichelberger:
I added a test case that fails without [the patch](https://github.com/Enigmatis/graphql-java-annotations/pull/301/commits/2f5b586a10ef5c8cd850b0014d09902a369bd6bb)

jkevan (approved):
Hi Geichelberger,

Thanks for your contribution! Great job on the GraphQL extensions name resolution. Merging now.

Best,
Kevan.

PR #310, opened 2025-03-17
Add support for partial updates

anttip:
Adds support for binding to `java.util.Optional` parameters. Binding to Optional parameters allows implementing partial updates. With a partial update the mutate call can update only some of the Optional fields without needing to repeat all the existing values of an object.

The mapping from an optional field in GraphQL schema that binds to `java.util.Optional` is the following:
1. value given to the field in GraphQL call is mapped into `java.util.Optional` of the underlying type
2. null given to the field in GraphQL call is mapped into `Optional.empty` to signal the removal of a possibly existing value of the field
3. field not included in the GraphQL call is mapped into null to signal `undefined` so any possibly existing value(s) should not be modified

### DIFF
--- a/src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
+++ b/src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
@@ -3,7 +3,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ * http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
@@ -20,10 +20,7 @@ import graphql.annotations.processor.typeFunctions.TypeFunction;
 import graphql.schema.*;
 
 import java.lang.reflect.*;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 import static graphql.annotations.processor.util.NamingKit.toGraphqlName;
 import static graphql.annotations.processor.util.PrefixesUtil.addPrefixToPropertyName;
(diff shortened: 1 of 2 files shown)


Fgerthoffert:
Thank you for the PR @anttip, 

After an initial look, this seems like a good idea, things are a bit busy on our end but we'll try to review that PR at some point in May (or earlier if we can).

Fgerthoffert:
Sorry for the delay, we'll try to review it by the end of June.

anttip:
Thanks for the update! Looking forward to the review. Meanwhile I can report that we have been using it without issues from our fork since creation. Hope it makes it to the project and is useful for others as well.

anttip:
@Fgerthoffert Any updates on this?

Fgerthoffert:
Apologies for the repeated delays, the past few months have been quite busy on our end. If everything goes as expected, we should review it within the next 2–3 weeks.

src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
```diff
@@ -148,8 +156,19 @@ private Object buildArg(Type p, GraphQLType graphQLType, Object arg) {
...
+                return Optional.of(list);
+            }
             return list;
+        } else if (p instanceof ParameterizedType) {
+            Type subType = ((ParameterizedType) p).getActualTypeArguments()[0];
+            Object val = buildArg(subType, graphQLType, arg);
+            // add Optional wrapper if needed
+            if (val != null && ((ParameterizedType) p).getRawType() == Optional.class) {
+                return Optional.of(val);
+            }
```
dgriffon: Small improvement
```suggestion
            if (((ParameterizedType) p).getRawType() == Optional.class) {
                return Optional.ofNullable(val);
            }
```

src/main/java/graphql/annotations/dataFetchers/MethodDataFetcher.java
```diff
@@ -120,6 +117,11 @@ private Object[] invocationArgs(DataFetchingEnvironment environment, ProcessingE
 
     private Object buildArg(Type p, GraphQLType graphQLType, Object arg) {
         if (arg == null) {
+            // for Optional parameters null should be returned as Optional.empty() to show a request for a null value
+            // and not including the parameter in the query at all should be returned as null to show "undefined" value / not set
+            if ((p instanceof ParameterizedType && ((ParameterizedType) p).getRawType() == Optional.class)) {
```
dgriffon: there is a not required bracket here:
```suggestion
              if (p instanceof ParameterizedType && ((ParameterizedType) p).getRawType() == Optional.class) {
```

dgriffon (changes requested):
Thank you for the contribution, globally everything is good, please fix the few suggestions, we'll be able to merge afterwards

anttip:
Thank you for the review and suggestions @dgriffon!

jkevan:
Thank you so much for your contribution! 🚀
Your work on adding support for partial updates is greatly appreciated. This enhancement will be valuable for the project and its users. Thank you for your time and effort in improving the codebase! 🙏

PR #317, opened 2026-05-11
2 allow extension of multiple interfaces for queries and mutations

richharms:


richharms:
My appologizes, I made a mistake opening this PR.

PR #319, opened 2026-05-12
Allow Inheritance of Multiple Classes and Interfaces for Query and Mutation Classes

richharms:
## Summary
Modifies functionality of two methods in AnnotationsSchemaCreator: queries and mutations, and allows the classes and interfaces that are passed in to have any super classes and extended interfaces to be examined for additional fields that should be added to the schema. This matches behavior of GraphQL objects.

This allows for better organization of source code (in my case, the interfaces containing queries and mutations have grown to be extremely unwieldy and need to be broken up by the major components in the application. This provides for that.

## Type
- [ ] Bug fix
- [X] Feature
- [ ] Docs
- [ ] Build/CI

## Checklist
- [X] Includes tests (or reason why not applicable)
- [ ] Updates documentation if needed
- [ ] Linked to issue: Fixes #____

## Notes for reviewers
Should be entirely backwards compatible, it was a drop-in replacement in my case.

### DIFF
--- a/src/main/java/graphql/annotations/processor/retrievers/GraphQLObjectInfoRetriever.java
+++ b/src/main/java/graphql/annotations/processor/retrievers/GraphQLObjectInfoRetriever.java
@@ -23,7 +23,9 @@ import java.lang.reflect.AnnotatedElement;
 import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.Comparator;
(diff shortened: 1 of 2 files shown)


Fgerthoffert:
Sorry for the delay in reviewing, we will try to provide a review by end of September 2026.

